PR #125, opened 2024-06-17
Get Credentials and Remove Role Fixes

mikewiebe:
**Summary of Changes:**

* Update to allows `'ndfc_device_username'` and `'ndfc_device_password'` credentials to be stored as environment variables or set directly under group_vars and retrieved properly.
* Refactor `remove` role to generate skip messages at the end to make them more noticeable/readable.

### DIFF
--- a/plugins/action/common/get_credentials.py
+++ b/plugins/action/common/get_credentials.py
@@ -34,21 +34,23 @@ display = Display()
 class ActionModule(ActionBase):
 
     def run(self, tmp=None, task_vars=None):
-        # self._supports_async = True
         results = super(ActionModule, self).run(tmp, task_vars)
-        results['failed'] = False
+        results['retrieve_failed'] = False
 
-        inv_list = self._task.args['inv_list']
-        username = task_vars.get('ndfc_device_username')
-        password = task_vars.get('ndfc_device_password')
+        key_username = 'ndfc_device_username'
+        key_password = 'ndfc_device_password'
+
+        ndfc_host_name = task_vars['inventory_hostname']
+        username = task_vars['hostvars'][ndfc_host_name].get(key_username, '')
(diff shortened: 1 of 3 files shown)


roles/dtc/common/tasks/ndfc_inventory.yml
```diff
@@ -48,3 +48,8 @@
     inv_list: "{{ inv_config }}"
   register: updated_inv_config
   no_log: true
+
+- name: Credential Retrieval Failed
```
mikewiebe, line 52: Added this because we don't see the error message generated by the action plugin to indicate `username` and `password` is not set because we have the `no_log: true` set.

plugins/action/common/get_credentials.py
```diff
@@ -34,21 +34,23 @@
...
-        # self._supports_async = True
         results = super(ActionModule, self).run(tmp, task_vars)
-        results['failed'] = False
+        results['retrieve_failed'] = False
 
-        inv_list = self._task.args['inv_list']
-        username = task_vars.get('ndfc_device_username')
-        password = task_vars.get('ndfc_device_password')
+        key_username = 'ndfc_device_username'
+        key_password = 'ndfc_device_password'
```
mtarking, line 41: Can we make these ```ndfc_switch_username``` and ```ndfc_switch_password```?

mtarking (changes requested):

roles/dtc/remove/tasks/main.yml
```diff
@@ -142,6 +114,34 @@
   when:
     - (inventory_delete_mode is defined) and (inventory_delete_mode is true|bool)
 
+- ansible.builtin.debug:
```
mikewiebe, line 117: Moved these to the end to make the skips more visible

mtarking (approved):

mtarking (approved):

PR #126, opened 2024-06-18
Update to use delegate_to localhost

### LINKED ISSUE #127: Tasks that are not run against NDFC should be delegated to the localhost
As the hosts is NDFC, task will use connection plugin to login to NDFC first before each task which will slow down the process. Please check all the tasks to make sure all tasks that should be run locally are delegated to the localhost.

mtarking:
Will resolve #127 
Update tasks, both NDFC and local to use delegate_to localhost


mikewiebe:
@mtarking Did you see improvement in the overall runtime?

mikewiebe (approved):

PR #128, opened 2024-06-18
[Experimental Fix] Fabric Version Check

mikewiebe:
This PR is experimental only.  Need to talk about how we want to handle these situations in the not so distant future.


roles/dtc/common/templates/ndfc_create_fabric_1222.j2
```diff
@@ -0,0 +1,188 @@
...
+{% endif %}
+{% endif %}
+{# ------------------------------------------------------ #}
+{# Underlay IPv4 Parameters                               #}
+{# ------------------------------------------------------ #}
+{% if (underlay.general.manual_underlay_allocation | default(defaults.vxlan.underlay.general.manual_underlay_allocation) | title) == 'False' %}
+  LOOPBACK0_IP_RANGE: {{ ipv4.underlay_routing_loopback_ip_range | default(defaults.vxlan.underlay.ipv4.underlay_routing_loopback_ip_range) }}
+  LOOPBACK1_IP_RANGE: {{ ipv4.underlay_vtep_loopback_ip_range | default(defaults.vxlan.underlay.ipv4.underlay_vtep_loopback_ip_range) }}
+  SUBNET_RANGE: {{ ipv4.underlay_subnet_ip_range | default(defaults.vxlan.underlay.ipv4.underlay_subnet_ip_range) }}
+  ANYCAST_RP_IP_RANGE: {{ ipv4.underlay_rp_loopback_ip_range | default(defaults.vxlan.underlay.ipv4.underlay_rp_loopback_ip_range) }}
```
mikewiebe, line 150: The following two lines represent the only difference between this file and the original `ndfc_create_fabric.j2` file.  We need to discuss alternative ways of versioning changes like this so we don't have to replicate the entire file + the changes.

roles/dtc/common/tasks/ndfc_fabric.yml
```diff
@@ -30,6 +30,14 @@
...
     dest: "{{ role_path }}/files/{{ fabric_file_name }}"
   delegate_to: localhost
+  when: ndfc_version.response.DATA.version is version('12.2.2', '<')
+
+- name: Build Fabric Creation Parameters From Template
+  ansible.builtin.template:
+    src: ndfc_create_fabric_1222.j2
+    dest: "{{ role_path }}/files/{{ fabric_file_name }}"
+  delegate_to: localhost
+  when: ndfc_version.response.DATA.version is version('12.2.2', '>=')
```
mtarking, line 40: Couldn't we do this check in the jinja file to limit the tasks (and skips) at runtime?

mtarking:
#203 should address this upon review and acceptance

mtarking:
#203 resolved this.

PR #138, opened 2024-06-25
Vrfs.yaml and default updates tests-integration folder

juburnet:
Found the issues in vrfs.yaml based on the robot tests.  
Went ahead and updated the copy of defaults (main.yaml) for the test_create folder.  We can later change this in a script or something, to keep an identical copy or link to the main default file in validate folder.


tests/integration/roles/test_create/defaults/main.yml
```diff
@@ -67,7 +67,7 @@ defaults:
             enable_bpdu_guard: true
           topology_switch_trunk_po_interface:
             description: 'NetAsCode Trunk PO Interface'
-            mtu: jumbo
+            mtu: jumbo
```
mtarking, line 70: Get rid of what looks like trailing whitespace

mtarking (approved):

tests/integration/roles/test_create/defaults/main.yml
```diff
@@ -67,7 +67,7 @@ defaults:
             enable_bpdu_guard: true
           topology_switch_trunk_po_interface:
```
mikewiebe, line 68: @juburnet can you check to see if we even need this defaults file?  What happens if we run the tests without it?

tests/integration/roles/test_create/defaults/main.yml
juburnet: will do.  This was an exact copy of the current defaults main.yaml.  I'll look at both files and remove any extra spaces.

mtarking: I fixed the trailing whitespace in the default main.yml in the validate role

tests/integration/roles/test_create/defaults/main.yml
juburnet: @mikewiebe - I was able to test more extensively today and we don't need the defaults file in the test_create folder.  When it exists in test_create, it does use it and it only uses this (it takes full priority over the validate defaults, and validate defaults is never referenced).    But when the defaults don't exist in test_create it uses the validate defaults.

We can remove the defaults folder and main.yml in the test_create folder.
I'll edit this PR

rrahimm (approved):

PR #140, opened 2024-06-26
Update readme for create and deploy roles

### LINKED ISSUE #135: Description of Create vs. Deploy role incorrect
Referring to "Understanding our Ansible roles" in the collection README/documentation, I think the description of the Create and Deploy role is not entirely correct. The documentation says "The deploy role is going to deploy those changes to the NDFC controller. [...] This is the role that is going to make the changes in the NDFC controller." 

However, I observe the Create role already applies the changes to the NDFC controller. The Deploy role applies the changes to the switches.

https://github.com/netascode/ansible-dc-vxlan
https://galaxy.ansible.com/ui/repo/published/cisco/nac_dc_vxlan/docs/

mtarking:
Update readme for create and deploy roles to resolve #135

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,9 +14,9 @@ This is achieved by creating YAML files that contain a pre-determined data schem
 
 > **Note**: For complete support and additional capabilities, Cisco provides a professional services capability under the Services as Code portfolio of services which can provide feature creation, end to end support and more.
 
-## Understanding our Ansible roles
+## Understanding our Ansible Roles
 
-### Validate role
+### Validate Role
 
 Role: [cisco.nac_dc_vxlan.validate](https://github.com/netascode/ansible-dc-vxlan/blob/develop/roles/validate/README.md)
 
@@ -24,19 +24,19 @@ The `validate` role ensures that the data model is correct and that the data mod
 
 As part of the VXLAN as Code service from Cisco, you will also be able to utilize the semantic validation feature to make sure that the data model matches the intended expected values. This is a powerful feature that allows you to ensure that the data model is correct before it is deployed to the network. Additonally the validate role allows creation of rules that can be used to prevent operators from making specific configurations that are not allowed in the network. These can be as simple as enforcing proper naming conventions to more complex rules for interconnectivity issues that should be avoided. These rules are coded in Python and can be constructed as part of the Services as Code offer. 
 
-### Create role
+### Create Role
 
 Role: [cisco.nac_dc_vxlan.dtc.create](https://github.com/netascode/ansible-dc-vxlan/blob/develop/roles/dtc/create/README.md)
 
 The `create` role builds all of the templates and variable parameters required to deploy the VXLAN fabric and creates fabric state in NDFC. The data model is converted into the proper templates required by the Ansible modules used to communicate with the NDFC controller and manage the fabric state. The `create` role has a dependency on the `validate` role.
 
-### Deploy role
+### Deploy Role
 
 Role: [cisco.nac_dc_vxlan.dtc.deploy](https://github.com/netascode/ansible-dc-vxlan/blob/develop/roles/dtc/deploy/README.md)
 
 The `deploy` role deploys the fabric state created using the Create role to the NDFC managed devices. The `deploy` role has a dependency on the `validate` role.
 
-### Remove role
+### Remove Role
 
 Role: [cisco.nac_dc_vxlan.dtc.remove](https://github.com/netascode/ansible-dc-vxlan/blob/develop/roles/dtc/remove/README.md)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,21 +20,21 @@ This is achieved by creating YAML files that contain a pre-determined data schem
...
 
 ### Create role
 
 Role: [cisco.nac_dc_vxlan.dtc.create](https://github.com/netascode/ansible-dc-vxlan/blob/develop/roles/dtc/create/README.md)
 
-This role is going to create all the templates and variable parameters that are going to be used in the deployment of the VXLAN fabric. This role converts the data model into the proper templates that are required by the Ansible module to be able to communicate with the NDFC controller.
+This role is going to use a templates and variable parameters from the data model to create and stage all the configuration for your VXLAN fabric. This role uses the data model in the proper templates to render the expected data that is required by the NDFC Ansible modules.
```
marehler, line 31: There are two steps. First, this role creates the configuration from the templates and variables. (Apparently by including nac_dc_vxlan/roles/dtc/common/tasks/main.yml, although I don't get how it is included.) Second, this roles applies the configuration to the NDFC, thereby creating the fabric on the NDFC. (By including nac_dc_vxlan/roles/dtc/create/tasks/main.yml.)

README.md
```diff
@@ -20,21 +20,21 @@ This is achieved by creating YAML files that contain a pre-determined data schem
...
 
 ### Deploy role
 
 Role: [cisco.nac_dc_vxlan.dtc.deploy](https://github.com/netascode/ansible-dc-vxlan/blob/develop/roles/dtc/deploy/README.md)
 
-The deploy role is going to deploy those changes to the NDFC controller. This role is going to take the templates and variable parameters that were created in the `create` role and deploy them to the NDFC controller. This is the role that is going to make the changes in the NDFC controller.
+The deploy role is going to recalculate and deploy the staged changes from the `create` role to your NDFC controller.
```
marehler, line 37: No, the deploy roles triggers NDFC to recalculate and deploy the configuration (which already has been applied to NDFC by the create role) to the switches.

marehler (changes requested):

README.md
mtarking: Hi @marehler, correct, but I don't think we need explicitly expand on all the dependencies and details under the covers as that's the point in the collection in taking that burden off the consumer. Essentially, one should not need to dive into the role. 

Also agree with your second sentence and that is what is meant by staging the configuration in NDFC. Create stages and does not push to the switches (except in a few one off scenarios to ensure certain things are in a working order for the collection to run). 

Let me know if that doesn't make sense or if we're crossing wires.

README.md
mtarking: Again, staged in fact means staging the configuration within NDFC, so I believe we're saying the same thing, no?

README.md
marehler: Maybe the term "stage" is a bit unclear. Besides it says "stage", not "stage in NDFC". Therefore I thought it meant just creating and preparing the configuration in Ansible but not pushing them yet to NDFC. How about rephrasing the first sentence to "[...] to create the configuration in NFDC"? Then it should be clear.

README.md
marehler: Currently it says "deploy to NDFC", but it's actually "deploy from NDFC to the switches". Therefore I thought it meant creating the configuration in NDFC what is done in the create role. I'd suggest to rephrase it like "The deploy role triggers NDFC to recalculate and deploy the configuration to the switches."

README.md
```diff
@@ -20,21 +20,21 @@ This is achieved by creating YAML files that contain a pre-determined data schem
 
 Role: [cisco.nac_dc_vxlan.validate](https://github.com/netascode/ansible-dc-vxlan/blob/develop/roles/validate/README.md)
 
-The validate role function is to ensure that the data model is correct, and that the data model is going to be able to be processed by the subsequent roles. The validate role is going to read all the files in the `host_vars` directory and create a single data model in memory for execution.
+The validate role has the function to ensure that the data model is correct, and that the data model is going to be able to be processed by the subsequent roles. The validate role is going to read all the files in the `host_vars` directory that is named after the inventory group defined in your inventory file and create a single data model in memory for execution.
```
mikewiebe, line 23: The documentation in our example repo is more accurate so let's use that language here.

README.md
mikewiebe: The documentation in our example repo is clearer so I propose we use that here as well.

README.md
mikewiebe: The documentation in our example repo is clearer so I propose we use that here as well.

README.md
marehler: Agree, that is clearer.

mtarking:
> Looks good now.

Thanks @marehler!

PR #142, opened 2024-06-26
Add Syslog to Fabric Template

mtarking:
Add syslog to the fabric template config settings.


mikewiebe (approved):

devegupt (approved):
@mtarking @mikewiebe This looks good to me, I tested it and works fine.

PR #145, opened 2024-07-01
Add Test Checks for NDFC Connectivity & Authorization

### LINKED ISSUE #104: Validate role should validate the connectivity to the fabric in the beginning of the playbook validate role
I suggest to check the communication and credentials to the ndfc, in the beginning of the playbook and validate role so that it will fail fast rather than doing all of the validation, prepare and build steps first.
"List fabric" could be used to test connectivity in the beginning.

mtarking:
This PR resolves #104 
Add test checks for NDFC & Switch connectivity & authorization in the validate role after schema and rule tests.

Need to determine if we want a control knob to turn these connectivity and auth tests on & off.


rrahimm:
@mtarking 

i believe we should have control knobs to turn these on/off.

There could be the scenario, where our solution will just be provided access to NDFC alone (as that is enough to run the solution). So, the tests towards Switches directly would fail.

So, we may have a default of Checking NDFC and not checking switches, and allow users to control with flags.

juburnet:
@mtarking I reviewed this.  Perhaps we should break this PR into 2 pieces.
It works well for the ndfc checks.  It succeeds when all is good, and it fails for timeout and for authentication issues. 

For the switch tests I'm unable to do a successful test because my laptop/Ansible doesn't have access to the switches directly.  I agree with previous comments that the control knob is required for switches.  For DTC we ideally want to test connectivity and credentials from the NDFC.  Not sure if it's possible, or if we want to basically do a discovery and see what's manageable.

If we split this PR, we could go ahead and implement the NDFC checks. Otherwise, we need the controls for switches, with a default of false, before we implement.

mtarking:
@juburnet please review again as I checked a few things with the NDFC check. Thanks.

juburnet (approved):
This works as expected.  Failures occur quickly for NDFC timeout or creds issues, and it works perfectly when connectivity to NDFC is good.
Thanks @mtarking

PR #147, opened 2024-07-08
Integration tests for the fabric creation

devegupt:
This PR includes full data profile for fabrics configuration as below:

1. fabric_full_small_ospf_multicast_asm_example
2. fabric_full_small_ospf_multicast_bidir_example
3. fabric_full_small_ospf_ingress_example
4. fabric_full_small_isis_multicast_asm_example
5. fabric_full_small_isis_multicast_bidir_example
6. fabric_full_small_isis_ingress_example

### DIFF
--- a/tests/integration/test_vxlan_small_isis_ingress.yml
+++ b/tests/integration/test_vxlan_small_isis_ingress.yml
@@ -19,33 +19,33 @@
 #
 # SPDX-License-Identifier: MIT
 
-all:
-  vars:
-    ansible_user: "admin"
-    ansible_python_interpreter: "~/.pyenv/shims/python"
-    ansible_httpapi_validate_certs: False
-    ansible_httpapi_use_ssl: True
+---
+# This playbook runs tests agains the data under tests/integration/host_vars/fabric_full_small_isis_ingress
 
-  children:
-    ndfc:
-      hosts:
-        fabric_empty:
-          ansible_host: << ndfc_ip >>
-        fabric_full_small:
-          ansible_host: < ndfc_ip >>
-        fabric_full_small_sha:
-          ansible_host: << ndfc_ip >> 
-        fabric_full_small_ospf_multicast_asm:
-          ansible_host: << ndfc_ip >>
-        fabric_full_small_ospf_multicast_bidir:
-          ansible_host: << ndfc_ip >>
-        fabric_full_small_ospf_ingress:
-          ansible_host: << ndfc_ip >>
-        fabric_full_small_isis_multicast_asm:
-          ansible_host: << ndfc_ip >>
-        fabric_full_small_isis_multicast_bidir:
-          ansible_host: << ndfc_ip >>
-        fabric_full_small_isis_ingress:
-          ansible_host: << ndfc_ip >>
-        fabric_full_large:
-          ansible_host: << ndfc_ip >>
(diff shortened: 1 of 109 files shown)


juburnet (approved):
@devegupt Reviewed all of these locally and they deployed correctly with the correct configs.  We can check with the team if they are fine to merge.

juburnet:
@devegupt  @mtarking  - to reduce a little duplication, lets remove fabric_full_small_ospf_multicast_asm_example since it already exists as part of fabric_full_small_example

mtarking:
I think it might be better to keep @devegupt's with the more descriptive name and replace the existing.

juburnet:
> I think it might be better to keep @devegupt's with the more descriptive name and replace the existing.

That would be fine.  We have been using fabric_full_small as the generic template, just like we have the generic fabric_full_large.  Either way would be fine.

tests/integration/test_vxlan_small_isis_ingress.yml
```diff
@@ -0,0 +1,51 @@
...
+#    ansible-playbook -i tests/integration/test_inventory.yml tests/integration/test_vxlan_small_isis_ingress.yml
+
+# Reset Fabric
+- hosts: fabric_empty
+  gather_facts: no
+
+  roles:
+    - test_remove
+
+# Create and Deploy Fabric Configuration
```
mikewiebe, line 39: Instead of duplicating the main playbook for every fabric type, I suggest we try the following approach instead.

Create a single playbook where the `Create and Deploy Fabric Configuration` section takes an ansible variable.

```yaml
# Create and Deploy Fabric Configuration
- hosts: '{{ hostvars_fabric_name }}'
  gather_facts: no

  roles:
    - test_create
```

Then I believe you should be able to run it like this and pick the fabric type you want to validate

```bash
ansible-playbook -i hosts test_inventory.yml -e 'hostvars_fabric_name=fabric_full_small_isis_ingress'
```

mikewiebe: If this works we can update the main playbook with the list of fabric profiles we support

devegupt:
> > I think it might be better to keep @devegupt's with the more descriptive name and replace the existing.
> 
> That would be fine. We have been using fabric_full_small as the generic template, just like we have the generic fabric_full_large. Either way would be fine.

@juburnet @mtarking I have removed the fabric_full_small_example fabric profile including all the references related to this in the test_inventory.yml and test_update_model_data_yml. Since we have the same configuration in fabric_full_small_ospf_multicast_asm.

tests/integration/test_vxlan_small_isis_ingress.yml
devegupt: @mikewiebe sounds good to me. I have created a generic playbook "test_vxlan.yml" with ansible variable to pass the fabric profile name, I tested and it works well. So I have updated the PR with removing all other playbooks related to different fabric profiles. I also added detailed comment having list of the supported profiles and Ansible command example to deploy different profiles. Please review and let me know your inputs.

  e.g. ansible-playbook -i test_inventory.yml test_vxlan.yml -e 'fabric_profile=fabric_full_small_isis_multicast_asm'

mikewiebe (approved):

PR #164, opened 2024-07-23
Add Fabric STP

mtarking:
Add Fabric STP to fabric j2 template.
Refactor fabric template order to better group sections.
Refactor ndfc util macro to make more generic.
Add a rule to ensure specific STP params are used with either rpvst+ or mst.


roles/validate/files/rules/required_rules/201_global_spanning_tree.py
```diff
@@ -0,0 +1,35 @@
...
+                        )
+
+                    if root_bridge_protocol == "rpvst+" and not vlan_range:
+                        results.append(
+                            "vxlan.global.spanning_tree.vlan_range must be set when the "
+                            "spanning tree vxlan.global.spanning_tree.root_bridge_protocol is set to rpvst+."
+                        )
+
+                    if root_bridge_protocol == "mst" and not mst_instance_range:
+                        results.append(
```
mikewiebe, line 30: Seems like this message should be similar to the message above on line `25,26`.

`vxlan.global.spanning_tree.mst_instance_range must be set when the spanning tree vxlan.global.spanning_tree.root_bridge_protocol is set to mst`

mtarking: I removed the latter two rule checks as they were unnecessary since we will pull from defaults if the user doesn't specify anything.

mikewiebe (approved):

mikewiebe (approved):

PR #172, opened 2024-08-02
Diff Run Framework

mikewiebe:
Summary:

This PR adds the framework to selectively run sections based on changes to the model data.  This significantly speeds up the run time when there are no changes and provides incremental improvements when only a subset of data changes are applied across the various sections.

The main logic driving these changes consist of two main parts:
  * A run_map that keeps track of which sections are completed as part of each run.
  * A diff action plugin that checks for differences in the files that are generated under `roles/dtc/common/files` between the current and previous runs.

These changes introduce three new action plugins:
* `plugins/action/common/read_run_map.py` - Reads changes to the `run_map.yaml` file under `roles/validate/files` which keeps track of each section that completed during the previous run.
* `plugins/action/common/run_map.py` - Writes changes to the `run_map.yaml` file under `roles/validate/files` which records each section that completes during the current run.
* `plugins/action/dtc/diff_model_changes.py` - Calculates the difference between files under `roles/dtc/common/files` from the previous run and current run.

This PR also introduces a new control parameter called `force_run_all` to force all sections to run and can be set under group_vars

Sections will be selectively run based on changes to the data model when the following is true:
* The `force_run_all` flag is set to `false`
* All sections in the run_map are set to `true` from the previous run


roles/validate/files/run_map.yml
```diff
@@ -0,0 +1,5 @@
+role_create_completed: true
```
mikewiebe, line 1: Add to gitignore

mikewiebe, line 1: Add `this file was autogenerated` note

juburnet (changes requested):
See notes for the parameter addition.  All other testing was successful.  I was able to add a vrf and remove the vrf and only those changes were made.  I'll do additional testing for interfaces, etc.

README.md
```diff
@@ -315,6 +315,24 @@ Example: Selectively Run `cisco.nac_dc_vxlan.create` role alone
...
+  * Control flag `force_run_all` under group_vars is set to `true`.
+  * When using ansible tags to control execution.
+  * When one of the following roles failed to complete on the previous run.
+    * `cisco.nac_dc_vxlan.validate`
+    * `cisco.nac_dc_vxlan.create`
+    * `cisco.nac_dc_vxlan.deploy`
+    * `cisco.nac_dc_vxlan.remove`
+
+  If any of these conditions is true then all roles/sections will be run.
+
```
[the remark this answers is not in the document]
mikewiebe: Good catch @juburnet.  I should also be able to add a default value so that if it's not defined the playbooks still executes with it set to `true`.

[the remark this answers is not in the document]
mikewiebe: I see the issue.  It's because I have it in a debug message and I am not checking to see if it exists first.  I will wrap that specific debug in a conditional so it does not have to be defined

mikewiebe:
Resolve this issue (https://github.com/netascode/ansible-dc-vxlan/issues/177) as part of this PR.

mtarking (approved):

roles/dtc/common/tasks/sub_main.yml
```diff
@@ -157,4 +143,32 @@
...
+      - "+     Interface Trunk PO Changes Detected -   [ {{ changes_detected_interface_trunk_po }} ]"
+      - "+     Sub Interface Routed Changes Detected - [ {{ changes_detected_sub_interface_routed }} ]"
+      - "+     ----- All Interfaces -----"
+      - "+     All Interfaces Changes Detected -       [ {{ changes_detected_interfaces }} ]"
+      - "+     ----- All Interfaces -----"
+      - "+     VRFs Changes Detected -                 [ {{ changes_detected_vrfs }} ]"
+      - "+     Networks Changes Detected -             [ {{ changes_detected_networks }} ]"
+      - "+     ----- Run Map -----"
+      - "+     Run Map Diff Run -                      [ {{ run_map_read_result.diff_run }} ]"
+      - " +    Force Run Flag  -                       [ {{ force_run_all }} ]"
```
juburnet, line 173: @mikewiebe When force_run_all is not defined in group_vars, it is erroring out for the playbook.  I don't see the default value defined in this PR.

juburnet (changes requested):

roles/dtc/common/tasks/sub_main.yml
mikewiebe: @juburnet Good catch.  It was a new file and I forgot to add it to the diffs.  Added now.

https://github.com/netascode/ansible-dc-vxlan/pull/172/files#diff-7584e1953f921f9c0b101993010aaae7ac16d33d25c3d5a2ceed4d693b5af351

juburnet (approved):
Everything works as expected.  The defaults also work, and are able to be overridden in the global_vars.  Ready to merge

PR #185, opened 2024-09-06
Reduce amount of data used during discovery device to speed up the task

ccoueffe:
Context: Testing on fabric having :
- 140 Leaves
- 1000 Vlans

Updating API to get Policies and filter with template equal to host_11_1 outside the loop in the task "Build Payload for NDFC Switch Hostname Policy Update"

This reduce the amount of data and improve time in the loop.

Shrink the data used on the task "Get Hostname for {{ item }} from Data Model" in the build_hostname_policy_payload to limit the data with only switches.


juburnet:
According to Charly, with this many Leaves and Vlans the current code fails deployment.  This fix is required to allow deployment successful

roles/dtc/create/tasks/devices_discovery.yml
```diff
@@ -29,26 +29,31 @@
...
-  ansible.builtin.set_fact:
-    md_serial_numbers: "{{ md_serial_numbers_list | join('%2C') }}"
-  delegate_to: localhost
-
-- name: Get NDFC Fabric Default Switch Policies
+- name: Get NDFC Fabric Default Switch Policies with temmplate name host_11_1
   cisco.dcnm.dcnm_rest:
     method: GET
-    path: "/appcenter/cisco/ndfc/api/v1/lan-fabric/rest/control/policies/switches?serialNumber={{ md_serial_numbers }}"
+    path: "/appcenter/cisco/ndfc/api/v1/lan-fabric/rest/control/policies/pagination?filter=templateName==host_11_1&fabricName={{ MD.vxlan.global.name }}"
```
mtarking, line 44: We cannot use this API as it's an internal-only API to NDFC and not publicly documented. It has potential to change without notice nor tracking thus not a good development practice to use. @dsx1123 @mikewiebe

mtarking (changes requested):

mtarking:
Resolved by #190

PR #187, opened 2024-09-13
Update task to account for login domain

mtarking:
This is to support different domains in current and future NDFC releases.


devegupt (approved):
Looks good and executed a test successfully.

juburnet (approved):
Approved

PR #189, opened 2024-09-15
Prepend fabric name

### LINKED ISSUE #184: Fix run_map for diff_run framework
The diff_run framework currently creates a run_map file under `collections/ansible_collections/cisco/nac_dc_vxlan/roles/validate/files/run_map.yml` which will work for a single fabric but this file needs to prepend the name of the fabric to avoid namespace collisions.

mtarking:
Prepend fabric name to prevent over-writing incorrect file and fabric map.

Resolves #184

### DIFF
--- a/plugins/action/common/read_run_map.py
+++ b/plugins/action/common/read_run_map.py
@@ -39,7 +39,8 @@ class ActionModule(ActionBase):
         results = super(ActionModule, self).run(tmp, task_vars)
         results['diff_run'] = True
 
-        fabric_name = task_vars['inventory_hostname']
+        model_data = task_vars['model_data']['data']
+        fabric_name = model_data["vxlan"]["global"]["name"]
 
         if 'dtc' in task_vars['role_path']:
             common_role_path = os.path.dirname(task_vars['role_path'])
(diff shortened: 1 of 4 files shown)


plugins/action/common/read_run_map.py
```diff
@@ -39,13 +39,15 @@ def run(self, tmp=None, task_vars=None):
         results = super(ActionModule, self).run(tmp, task_vars)
         results['diff_run'] = True
 
+        fabric_name = task_vars['inventory_hostname']
```
devegupt, line 42: I executed a few tests with updating the fabric name under vxlan.global.name, I couldn't see the new run_map file with fabric name prefix. As per this line in the code, I understand we are actually prepending inventory name and not the fabric name with run_map.yml file.

devegupt (commented):
Added a comment in read_run_map.py.

plugins/action/common/read_run_map.py
mtarking: @devegupt interesting point you raise; lets discuss in the standup.

juburnet (changes requested):

plugins/action/common/read_run_map.py
mtarking: Right, that's expected with that commit. Before modifying the plugin, the test was to make sure things indeed still worked correctly. I'll modify the code now.

PR #190, opened 2024-09-18
Refactor Build & Set Hostname for Switch Inventory

### LINKED ISSUE #188: Speed up Setting Hostnames
Need to explore options (API & path queries) for speeding up setting the hostname tasks.
Ideally reduce Ansible tasks and move into native Python.

mtarking:
Resolves #188 
Refactor build & set hostname for switch inventory using a Python plugin to speed up execution in larger environments.


plugins/action/dtc/build_hostname_policy_payload.py
```diff
@@ -0,0 +1,57 @@
...
+                task_vars=task_vars,
+                tmp=tmp
+            )
+            policy_match = next((item for item in policy_data["response"]["DATA"] if item["templateName"] == "host_11_1"))
+            if policy_match["nvPairs"]["SWITCH_NAME"] != switch["name"]:
+                policy_match["nvPairs"]["SWITCH_NAME"] = switch["name"]
+                policy_payload.append(policy_match)
+
+        results['policy_payload'] = policy_payload
+        results['policy_ids'] = "%2C".join([str(policy["id"]) for policy in policy_payload])
```
dsx1123, line 55: suggest to return list instead of comma separated string, join the list in the playbook

dsx1123 (changes requested):

plugins/action/dtc/build_hostname_policy_payload.py
mtarking: returned just a list and re-added a task to perform the conversion to string list.
also added a result changed knob.

plugins/action/dtc/build_hostname_policy_payload.py
```diff
@@ -0,0 +1,60 @@
...
+from ansible.plugins.action import ActionBase
+
+
+class ActionModule(ActionBase):
+
+    def run(self, tmp=None, task_vars=None):
+        results = super(ActionModule, self).run(tmp, task_vars)
+        results['changed'] = False
+
+        model_data = self._task.args["model_data"]
```
dsx1123, line 36: @mtarking after second thought, I feel like we might need similar function in the future to update other policies that are created by the NDFC, will it be a good idea to make it generic? The input should be a list of switch serial numbers and the output is a list of dict:
```json
{
   "{serial_number}": "POLICY-XXXXX",
}
```

mtarking: Discussed verbally and agreed.

dsx1123 (approved):

PR #193, opened 2024-10-04
inital changes to support pre provisioned switches and links

mthurstocisco:
WIP for a number of changes required to support the build we will need for Barclays going forward, draft branch just for everyone's visibility.


roles/dtc/common/templates/ndfc_create_fabric.j2
```diff
@@ -102,6 +102,8 @@
 {# ------------------------------------------------------ #}
 {# Underlay General Parameters                            #}
 {# ------------------------------------------------------ #}
+  ENABLE_SGT: false
```
mthurstocisco, line 105: needed for 12.2.2

PR #196, opened 2024-10-08
Refactor fabric template to fix jinja set bug

mtarking:
@mikewiebe refactoring fabric jinja template to work around inconsistency in jinja.


mikewiebe:
@mtarking Did this fix the issue you were seeing?

mtarking:
> @mtarking Did this fix the issue you were seeing?

@mikewiebe yes it did. For this update, i just limited the number of sets and used the top-level of the model inline and if that's not found then the default is picked up. For underlay specifically, I did leave a set because of the number of items underneath it but if not found, then default to an empty dict. I'm all for some more additional testing, but things worked fine with fabric provisioning when I retested.

mikewiebe (approved):

PR #198, opened 2024-10-09
Updates to remove ignores and replace with module placeholder

mtarking:
Updates to remove ignores and replace with module placeholder properly.


mtarking:
New, more updated PR to follow.

PR #225, opened 2024-11-07
Fix tags

mikewiebe:
## Related Issue(s)

This simplifies how we tag our roles and fixes a problem where we cannot run the tags in isolation.

The new role `connectivity_check` needs to include tags for `create, deploy and remove` to make sure it gets run when we specify those roles

## Related Collection Role

* [x] cisco.nac_dc_vxlan.dtc.deploy
* [x] cisco.nac_dc_vxlan.dtc.remove
* [ ] other

## Related Data Model Element

* [ ] vxlan.global
* [ ] vxlan.topology
* [ ] vxlan.underlay
* [ ] vxlan.overlay_services
* [ ] vxlan.policy
* [ ] defaults.vxlan
* [ ] other

## Test Notes

Using the following playbook I tested the following scenarios

* All roles run together
* Roles run in isolation
* Roles run selectively using tags

```yaml
---
# This is the main entry point playbook for calling the various
# roles in this collection.
- hosts: nac-ndfc1
  any_errors_fatal: true
  gather_facts: no
  
  roles:
    # Prepare service model for all subsequent roles
    # Note - The validate role is run automatically as a prerequisite to the create, deploy, and remove roles.
    #
    # - role: cisco.nac_dc_vxlan.validate
    # -----------------------
    # DataCenter Roles
    #   Role: cisco.netascode_dc_vxlan.dtc manages direct to controller NDFC workflows
    #
      tags: 'role_create'

      tags: 'role_deploy'

    - role: cisco.nac_dc_vxlan.dtc.remove
      tags: 'role_remove'
```

## Checklist

### DIFF
--- a/roles/dtc/common/tasks/main.yml
+++ b/roles/dtc/common/tasks/main.yml
@@ -23,4 +23,4 @@
 
 - name: Import Role Tasks
   ansible.builtin.import_tasks: sub_main.yml
-  # tags: "{{ nac_tags.all }}" # Tags defined in roles/common_global/vars/main.yml
+  tags: "{{ nac_tags.common_role }}" # Tags defined in roles/common_global/vars/main.yml
--- a/roles/dtc/common/tasks/main.yml
+++ b/roles/dtc/common/tasks/main.yml
@@ -23,4 +23,4 @@
 
 - name: Import Role Tasks
   ansible.builtin.import_tasks: sub_main.yml
-  tags: "{{ nac_tags.all }}" # Tags defined in roles/common_global/vars/main.yml
+  # tags: "{{ nac_tags.all }}" # Tags defined in roles/common_global/vars/main.yml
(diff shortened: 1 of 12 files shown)


roles/dtc/common/tasks/main.yml
```diff
@@ -23,4 +23,4 @@
 
 - name: Import Role Tasks
   ansible.builtin.import_tasks: sub_main.yml
-  tags: "{{ nac_tags.all }}" # Tags defined in roles/common_global/vars/main.yml
+  # tags: "{{ nac_tags.all }}" # Tags defined in roles/common_global/vars/main.yml
```
mikewiebe, line 26: I commented this out because it causes common to run when we try and run the validate role by itself.  Common should not run in this case and since common is a "hidden" infra role we don't call it in isolation so we don't need to tag it

mtarking (approved):

Issue #158, opened 2024-07-17
Policy: ACLs


mtarking:

devegupt:
We can refer to the schema defined in this PR https://wwwin-github.cisco.com/netascode/nac-vxlan/pull/58/files

devegupt:
Uploading the draft merged schema and data yaml file having the route map, acl, prefix list data model. 
[updated_schema_datafile_draft.zip](https://github.com/user-attachments/files/17379672/updated_schema_datafile_draft.zip)

devegupt:
@mtarking, we checked this in the sync up call and schema items are missing for this feature (policy: ACL).

mtarking:
@devegupt Correct. My draft is for route-maps, prefix-lists, communities, etc as they can be used in the former.

juburnet:
This was merged in [Update to support route_control #248](https://github.com/netascode/ansible-dc-vxlan/pull/248)

Issue #157, opened 2024-07-17
Policy: Prefix-Lists


mtarking:

devegupt:
Uploading the draft merged schema and data yaml file having the route map, acl, prefix list data model.
[updated_schema_datafile_draft.zip](https://github.com/user-attachments/files/17379675/updated_schema_datafile_draft.zip)

juburnet:
This was merged in [Update to support route_control #248](https://github.com/netascode/ansible-dc-vxlan/pull/248)

Issue #156, opened 2024-07-17
Policy: Route-Maps


mtarking:

juburnet:
@mtarking - I'll start on the Jinja for this, and look at creating similar structure to VRF-lite for the rest of the implementation.

devegupt:
Uploading the draft merged schema and data yaml file having the route map, acl, prefix list data model.
[updated_schema_datafile_draft.zip](https://github.com/user-attachments/files/17379678/updated_schema_datafile_draft.zip)

juburnet:
This was merged in [Update to support route_control #248](https://github.com/netascode/ansible-dc-vxlan/pull/248)

PR #248, opened 2024-12-10
Update to support route_control

ccoueffe:
## Related Issue(s)

Fix the following issues:
- #158 
- #157
- #156 

## Related Collection Role

* [ ] other

## Related Data Model Element

* [ ] vxlan.global
* [ ] vxlan.topology
* [ ] vxlan.underlay
* [ ] vxlan.overlay_services
* [ ] vxlan.policy
* [ ] defaults.vxlan
* [ ] other

## Proposed Changes

This change includes templates to generate policies for:
- Route-map
- Prefix-list (IPv4 and IPv6)
- Communities (Standard and Extended)
- IP AS-Path
- MAC-List
- ACLs (IPv4 and IPv6)
- Time-Range for ACLs
- Object-Groups (IPv4 and IPv6) for ACLs

## Cisco NDFC Version

Tested with ND3.2(1i)

## Checklist


roles/dtc/common/templates/ndfc_route_control.j2
```diff
@@ -0,0 +1,26 @@
+{% for policy_types in group_item %}
+  {% if "route_maps" in policy_types %}
+    {% include "ndfc_route_map.j2" %}
```
mtarking, line 3: @ccoueffe this will apply to all the other includes as well, but can you please group the different include j2 files for this into a folder under ```templates``` and call it ```ndfc_route_control``` and then update the include lines in this file? Thanks!

mtarking (changes requested):

plugins/action/common/prepare_plugins/prep_108_route_control.py
```diff
@@ -0,0 +1,298 @@
...
+# COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+# IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+# CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+#
+# SPDX-License-Identifier: MIT
+
+import re
+from jinja2 import ChainableUndefined, Environment, FileSystemLoader
+from ansible_collections.ansible.utils.plugins.filter import ipaddr
+from ansible_collections.ansible.utils.plugins.filter import hwaddr
```
mtarking, line 25: @ccoueffe checkout the PEP8 guidelines [here](https://peps.python.org/pep-0008/#imports) for these; you can do on one line

plugins/action/common/prepare_plugins/prep_108_route_control.py
```diff
@@ -0,0 +1,298 @@
...
+                new_interfaces.append("Ethernet" + intf_ethernetdot1q[1])
+            elif intf_loopback:
+                new_interfaces.append("loopback" + intf_loopback[1])
+            elif intf_portchannel:
+                new_interfaces.append("port-channel" + intf_portchannel[1])
+            elif intf_portchanneldot1q:
+                new_interfaces.append("port-channel" + intf_portchanneldot1q[1])
+            elif intf_null:
+                new_interfaces.append("Null" + intf_null[1])
+        return new_interfaces
```
mtarking, line 298: Why are you having to do this extra functional code vs just ignoring case where you call this helper function?

plugins/action/common/prepare_plugins/prep_108_route_control.py
```diff
@@ -0,0 +1,298 @@
...
+                                        ],
+                                        "priority": 500
+                                    }
+                                    model_data["vxlan"]["policy"]["groups"].append(new_group)
+
+            model_data = hostname_to_ip_mapping(model_data)
+        self.kwargs['results']['model_extended'] = model_data
+        return self.kwargs['results']
+
+    def convert_port_number(self, port_number, protocol):
```
mtarking, line 193: Is there a reason the model didn't take the well-known cli? Asking because I don't see any of this in the data model. Can you provide a pointer here or via unicast?

roles/dtc/common/tasks/sub_main.yml
```diff
@@ -180,3 +180,5 @@
       - "+     Run Map Diff Run -                      [ {{ run_map_read_result.diff_run }} ]"
       - "+     Force Run Flag  -                       [ {{ force_run_all }} ]"
       - "----------------------------------------------------------------"
+
+- ansible.builtin.meta: end_play
```
mtarking, line 184: Need to remove.

roles/dtc/common/templates/ndfc_route_control/ndfc_route_control.j2
```diff
@@ -0,0 +1,26 @@
+{% for policy_types in group_item %}
```
mtarking, line 1: If this is the main, top-level route_control template, please let it sit at the root of the ```template``` directory

mtarking (changes requested):

plugins/action/common/prepare_plugins/prep_108_route_control.py
ccoueffe: Done. I merged both import with one line.

roles/dtc/common/tasks/sub_main.yml
ccoueffe: removed

roles/dtc/common/templates/ndfc_route_control/ndfc_route_control.j2
ccoueffe: moved

plugins/action/common/prepare_plugins/prep_108_route_control.py
```diff
@@ -0,0 +1,297 @@
...
+            undefined=ChainableUndefined,
+            lstrip_blocks=True,
+            trim_blocks=True,
+        )
+
+        env.filters["ipaddr"] = ipaddr.ipaddr
+        env.filters["hwaddr"] = hwaddr.hwaddr
+        template = env.get_template(template_filename)
+
+        if "overlay_extensions" in model_data["vxlan"]:
```
mikewiebe, line 57: We should be able to significantly simplify this logic and remove the deep nesting by making use of data_model_key_check.

For example you can use it to walk the model_data tree with a set of keys and inspect to see if there is data at a particular level.  If you want to see if there is data under `["vxlan"]["overlay_extensions"]["route_control"]["route_maps"]` without all of the looping you could do this.

```python
from ....plugin_utils.helper_functions import data_model_key_check

parent_keys = ['vxlan', 'overlay_extensions', 'route_control', 'route_maps']
dm_check = data_model_key_check(model_data, parent_keys)
```

dm_check will contain the following dictionaries

```
{'keys_found': [], 'keys_not_found': [], 'keys_data': [], 'keys_no_data': []}
```

* 'keys_found' - indicates we found the keys in the model data but not necessarily values for those keys
* 'keys_not_foind' - indicates we did not find the keys in the model data
* 'keys_data' - indicates we found the keys and values/data associated with the keys
* 'keys_no_data' - indicates we found keys but no values/data associated with the keys

mikewiebe: Our prep_001 plugin has good examples of how its used to avoid deep nesting

https://github.com/netascode/ansible-dc-vxlan/blob/a9e6d181e79c430a21504aa1ad6fa53f1ef7a8dc/plugins/action/common/prepare_plugins/prep_001_list_defaults.py#L76

mikewiebe: Also might want to consider creating this function in the helper_functions library to safely pull data from the model

https://github.com/netascode/ansible-dc-vxlan/blob/a9e6d181e79c430a21504aa1ad6fa53f1ef7a8dc/roles/validate/files/rules/required_rules/401_overlay_services_cross_reference.py#L72

roles/dtc/common/templates/ndfc_route_control/ndfc_ip_access_list.j2
```diff
@@ -0,0 +1,357 @@
+! This NDFC Route_Control_ACLs config data structure is auto-generated
+! DO NOT EDIT MANUALLY
+!
+!
+{% if policy_types["ipv4_access_lists"] is defined %}
```
mikewiebe, line 5: As discussed please see if it makes sense to simplify this further by including additional child templates and split along technology lines (v4 and v6) for instance.

roles/dtc/common/templates/ndfc_route_control/ndfc_route_map.j2
```diff
@@ -0,0 +1,402 @@
+! This NDFC Route_Control_Route_maps config data structure is auto-generated
+! DO NOT EDIT MANUALLY
+!
```
mikewiebe, line 3: Same question here about separation into child include files for debugability / maintainability

plugins/action/common/prepare_plugins/prep_108_route_control.py
```diff
@@ -0,0 +1,318 @@
...
+        )
+
+        env.filters["ipaddr"] = ipaddr.ipaddr
+        env.filters["hwaddr"] = hwaddr.hwaddr
+        template = env.get_template(template_filename)
+
+        parent_keys = ['vxlan', 'overlay_extensions', 'route_control', 'route_maps']
+        dm_check = data_model_key_check(model_data, parent_keys)
+
+        # Check Route-Maps
```
mikewiebe, line 60: Ok this is easier to read/understand now so thanks for refactoring.  Would it be safe to say that `self.route_maps` and `self.ipv[4|6]_access_list` are actually updating and transforming model data?  If so can se update the comments to indicate we are actually updating and transforming the data and add `update_` to the function names to be more clear when calling the function.

plugins/action/common/prepare_plugins/prep_108_route_control.py
```diff
@@ -0,0 +1,318 @@
...
+                                if isinstance(option_set["ipv4"]["precedence"], int):
+                                    option_set["ipv4"]["precedence"] = precedence_translation[
+                                        option_set["ipv4"]["precedence"]]
+
+                            if ("ipv6" in option_set) and ("precedence" in option_set["ipv6"]):
+                                if isinstance(option_set["ipv6"]["precedence"], int):
+                                    option_set["ipv6"]["precedence"] = precedence_translation[
+                                        option_set["ipv6"]["precedence"]]
+
+    def ipv4_access_list(self, model_data):
```
mikewiebe, line 173: It looks like `ipv4_access_list` and `ipv6_access_list` functions are almost identical.  Could you just use one function instead of two and pass in a key (either `ipv4_access_lists`, or `ipv6_access_lists`?

roles/dtc/common/templates/ndfc_route_control/ndfc_route_map.j2
ccoueffe: Done

plugins/action/common/prepare_plugins/prep_108_route_control.py
ccoueffe: Done. Refactor with one function: update_ip_access_lists()

plugins/action/common/prepare_plugins/prep_108_route_control.py
ccoueffe: rename functions for route-maps and IP ACLs.
* update_route_maps()
* update_ip_access_lists()

plugins/action/common/prepare_plugins/prep_108_route_control.py
ccoueffe: Schema allows interfaces with different cases and short or full name. However in the CLI, `Ethernet` and `Null` interfaces have a Capital letter, which is not the case for `port-channel` and `loopback`.

This function reformat interface names with the full name and the good case.

Ex:
```ios
  route-map rm-interfaces permit 10
      description Match interfaces
      match interface Ethernet1/10 loopback100 Null0 port-channel100
```

```yaml

---
vxlan:
  overlay_extensions:
    route_control:
      route_maps:
        - name: rm-interfaces
          entries:
            - seq_number: 10
              operation: permit
              description: Match interfaces
              match:
                - interfaces:
                    - Ethernet1/10
                    - loopback100
                    - null0
                    - po100
      groups:
        - name: leaf1
          route_maps:
            - name: rm-interfaces
      switches:
        - name: netascode-leaf1
          groups: leaf1
```

mtarking: Using Jinja filters wouldn't it be easier to match on ```eth``` for example and use the title filter vs having to maintain a mapper? Same applies to Null. Also, same could be applied with matching ```po``` or ```port``` for port-channel and use the lower filter. Thoughts???

cc @mikewiebe

ccoueffe: Currently I updated on the python script and join all interfaces in Jinja:

Ex:
```
  match interface {{ match["interface"] | join(" ") }}
```

Should I move all in Jinja ?

plugins/action/common/prepare_plugins/prep_108_route_control.py
mikewiebe: Discussed this with Charly and team and I think it makes more sense to use the port_number in the model and then map this to the well-known CLI.  We know the port number won't change but the CLI in theory could change and so mapping it is probably more "future proof".  Also based on discussions with the team if we just send the port number it results in an "out of sync" condition with NDFD because it gets converted to the CLI.

mikewiebe:
Charly still needs to address a mapping case with ICMP before we can merge this PR.

juburnet:
@mikewiebe - this was updated for ICMP mapping.  Please review and merge

juburnet (approved):
Approved.  It was agreed on a previous call to keep Port numbers only and have the translation to avoid the out-of-sync

PR #267, opened 2025-01-24
Update prep_108_route_control.py

ccoueffe:
update script to avoid duplicate policy in the extended model: vxlan.policy.groups and vxlan.policy.policies.

Duplicate appears because we removed switch name in the policy description.

## Related Issue(s)

Remove duplicate in the extended data model for the following keys:
* vxlan.policy.policies
* vxlan.policy.groups

## Related Collection Role

* [ ] other

## Related Data Model Element

* [ ] vxlan.global
* [ ] vxlan.topology
* [ ] vxlan.underlay
* [ ] vxlan.overlay_services
* [ ] vxlan.policy
* [ ] defaults.vxlan
* [ ] other

## Checklist

### DIFF
--- a/plugins/action/common/prepare_plugins/prep_108_route_control.py
+++ b/plugins/action/common/prepare_plugins/prep_108_route_control.py
@@ -103,7 +103,8 @@ class PreparePlugin:
                                 }
                             }
 
-                            model_data["vxlan"]["policy"]["policies"].append(new_policy)
+                            if not any(policy['name'] == unique_name for policy in model_data["vxlan"]["policy"]["policies"]):
+                                model_data["vxlan"]["policy"]["policies"].append(new_policy)
 
                             if any(sw['name'] == switch['name'] for sw in model_data["vxlan"]["policy"]["switches"]):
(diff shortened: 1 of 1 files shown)


mikewiebe:
Was there a background discussion to remove the switch name from the policy?  I don't recall if there was a discussion.

ccoueffe:
@mikewiebe, at the beginning we create the policy name, which is the description in ND, like for VRF-Lite based on the Policy name + Switch  Name. When switches had the same policies we push the same content with different name.
We agree to remove switch_name to have the same policy name. We simplify and unified the policy name, when they are similar. Now the behaviour is similar to freeform policy.

mikewiebe:
> @mikewiebe, at the beginning we created the policy name, which is the description in ND, like for VRF-Lite based on the Policy name + Switch Name. When switches had the same policies we push the same content with different name. We agree to remove switch_name to have the same policy name. We simplify and unified the policy name, when they are similar. Now the behaviour is similar to freeform policy.

Ah, that's right.  Thanks for the reminder.

mikewiebe (approved):

PR #563, opened 2025-08-22
Add support for ipv6 route map under vrf

### LINKED ISSUE #540: Support for VRF IPv6 Redistribute Direct Route Map

cisco.nac_dc_vxlan.dtc.create

### New feature description:

My customer uses different VRF redistribute route maps for IPv4 and IPv6. NaC VXLAN supports Redistribute Direct Route Map (redist_direct_routemap) but does yet not support IPv6 Redistribute Direct Route Map (ipv6_redist_direct_routemap).

n/a

vxlan.overlay

Yes, ipv6_redist_direct_routemap attribute needs to be added to vxlan.overlay.vrfs.

n/a

n/a

### Ansible Version

```shell

```

```shell
Collection                               Version  
---------------------------------------- ---------
ansible.netcommon                        7.1.0    
ansible.posix                            2.0.0    
ansible.utils                            5.1.2    
cisco.dcnm                               3.8.1    
cisco.nac_dc_vxlan                       0.4.3-dev
cisco.nxos                               9.4.0    
community.general                        10.1.0
```

### Cisco NDFC Version

3.2.1(i)

### Cisco NX-OS Version

10.5(3)

mkilar123:
PR is to add support for IPv6 route-map used for redistribution of directly connected routes

## Related Issue(s)
Fixes #540 

## Related Collection Role

* [ ] other

## Related Data Model Element

* [ ] vxlan.fabric
* [ ] vxlan.global
* [ ] vxlan.topology
* [ ] vxlan.underlay
* [x] vxlan.overlay
* [ ] vxlan.policy
* [ ] vxlan.multisite
* [ ] defaults.vxlan
* [ ] other

## Test Notes
Tested creation of vrfs with:
- default rm name specified,
- non-default rm name specified,
- rm name skipped in yaml files

I deleted conditional check under "[msd_child_fabric_vrf.j2](https://github.com/netascode/ansible-dc-vxlan/compare/develop...mkilar123:ansible-dc-vxlan:v6_redist_rmap?expand=1#diff-edac9fb9c809991da7f3927463c4bc6bc93416e07154ee5de648a8ab65fcc73c)
The reason is that under a child fabric the name of the route-map is "greyed out" and it always contains the value specified under MSD fabric level, so there is no option for this value to be different 0 - i basically treat the ipv6 route-map the same way as ipv4 one is.

## Cisco NDFC Version
3.2.1i

## Checklist

### DIFF
--- a/roles/dtc/common/templates/ndfc_vrfs/dc_vxlan_fabric/dc_vxlan_fabric_vrfs.j2
+++ b/roles/dtc/common/templates/ndfc_vrfs/dc_vxlan_fabric/dc_vxlan_fabric_vrfs.j2
@@ -54,6 +54,7 @@
   underlay_mcast_ip: {{ vrf['underlay_mcast_ip'] | default(omit) }}
 {% endif %}
   redist_direct_rmap: {{ vrf['redist_direct_routemap'] | default(defaults.vxlan.overlay.vrfs.redist_direct_routemap) }}
+  v6_redist_direct_rmap: {{ vrf['ipv6_redist_direct_routemap'] | default(defaults.vxlan.overlay.vrfs.ipv6_redist_direct_routemap) }}
 {# ------------------------------------------------------ #}
 {# Attach Group Section                                   #}
 {# ------------------------------------------------------ #}
--- a/roles/dtc/common/templates/ndfc_vrfs/dc_vxlan_fabric/dc_vxlan_fabric_vrfs.j2
+++ b/roles/dtc/common/templates/ndfc_vrfs/dc_vxlan_fabric/dc_vxlan_fabric_vrfs.j2
@@ -54,7 +54,7 @@
   underlay_mcast_ip: {{ vrf['underlay_mcast_ip'] | default(omit) }}
 {% endif %}
   redist_direct_rmap: {{ vrf['redist_direct_routemap'] | default(defaults.vxlan.overlay.vrfs.redist_direct_routemap) }}
-  v6_redist_direct_rmap: {{ vrf['ipv6_redist_direct_routemap'] | default(defaults.vxlan.overlay.vrfs.ipv6_redist_direct_routemap) }}
+{#  v6_redist_direct_rmap: {{ vrf['ipv6_redist_direct_routemap'] | default(defaults.vxlan.overlay.vrfs.ipv6_redist_direct_routemap) }} #}
 {# ------------------------------------------------------ #}
 {# Attach Group Section                                   #}
 {# ------------------------------------------------------ #}
--- a/roles/dtc/common/templates/ndfc_vrfs/dc_vxlan_fabric/dc_vxlan_fabric_vrfs.j2
+++ b/roles/dtc/common/templates/ndfc_vrfs/dc_vxlan_fabric/dc_vxlan_fabric_vrfs.j2
@@ -54,7 +54,7 @@
   underlay_mcast_ip: {{ vrf['underlay_mcast_ip'] | default(omit) }}
 {% endif %}
   redist_direct_rmap: {{ vrf['redist_direct_routemap'] | default(defaults.vxlan.overlay.vrfs.redist_direct_routemap) }}
-{#  v6_redist_direct_rmap: {{ vrf['ipv6_redist_direct_routemap'] | default(defaults.vxlan.overlay.vrfs.ipv6_redist_direct_routemap) }} #}
+  v6_redist_direct_rmap: {{ vrf['ipv6_redist_direct_routemap'] | default(defaults.vxlan.overlay.vrfs.ipv6_redist_direct_routemap) }}
 {# ------------------------------------------------------ #}
 {# Attach Group Section                                   #}
 {# ------------------------------------------------------ #}
--- a/roles/dtc/common/templates/ndfc_vrfs/dc_vxlan_fabric/dc_vxlan_fabric_vrfs.j2
+++ b/roles/dtc/common/templates/ndfc_vrfs/dc_vxlan_fabric/dc_vxlan_fabric_vrfs.j2
@@ -54,7 +54,9 @@
   underlay_mcast_ip: {{ vrf['underlay_mcast_ip'] | default(omit) }}
 {% endif %}
   redist_direct_rmap: {{ vrf['redist_direct_routemap'] | default(defaults.vxlan.overlay.vrfs.redist_direct_routemap) }}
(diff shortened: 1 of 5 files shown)


roles/dtc/common/templates/ndfc_vrfs/msd_fabric/child_fabric/3.2/msd_child_fabric_vrf.j2
```diff
@@ -16,7 +16,7 @@
       "mtu": ndfc.mtu,
       "tag": ndfc.tag,
       "vrfRouteMap": ndfc.vrfRouteMap,
-      "v6VrfRouteMap": ndfc.vrfRouteMapV6 if ndfc.vrfRouteMapV6 is defined else "",
+      "v6VrfRouteMap": ndfc.v6VrfRouteMap,
```
mtarking, line 19: What happens when v6VrfRouteMap nvpair does not have a value in ND?

roles/validate/files/defaults.yml
```diff
@@ -257,6 +255,7 @@ factory_defaults:
         no_rp: false
         rp_external: false
         redist_direct_routemap: FABRIC-RMAP-REDIST-SUBNET
+        ipv6_redist_direct_routemap: FABRIC-RMAP-REDIST-SUBNET
```
mtarking, line 258: Was this tested in setting this default vs letting the controller default the rm?

roles/dtc/common/templates/ndfc_vrfs/dc_vxlan_fabric/dc_vxlan_fabric_vrfs.j2
```diff
@@ -54,6 +54,7 @@
   underlay_mcast_ip: {{ vrf['underlay_mcast_ip'] | default(omit) }}
 {% endif %}
   redist_direct_rmap: {{ vrf['redist_direct_routemap'] | default(defaults.vxlan.overlay.vrfs.redist_direct_routemap) }}
+  v6_redist_direct_rmap: {{ vrf['ipv6_redist_direct_routemap'] | default(defaults.vxlan.overlay.vrfs.ipv6_redist_direct_routemap) }}
```
mtarking, line 57: I don't think adding this works for 3.1 nor in conjunction with the default.

roles/dtc/common/templates/ndfc_vrfs/msd_fabric/child_fabric/3.2/msd_child_fabric_vrf.j2
mkilar123: @mtarking 
"v6VrfRouteMap" is implemented exactly the same way as "vrfRouteMap":
1. Both have the default value assigned at the dcnm_vrf module level (at base collection).
2. At the child fabric-level the value is not editable (it's grey-ed out). It can only be set at the MSD level.
3. In the situation when this is skipped at the yaml level, we have the defaults at both SOLUTION and BASE module - so it will get set to FABRIC-RMAP-REDIST-SUBNET - exactly the same way as it happens for "vrfRouteMap"

roles/validate/files/defaults.yml
mkilar123: @mtarking 
As described in the PR itself, for both individual fabric and the MSD i tested the three scenarios: 
-     default rm name specified,
-     non-default rm name specified,
-     rm name skipped in yaml files

roles/dtc/common/templates/ndfc_vrfs/dc_vxlan_fabric/dc_vxlan_fabric_vrfs.j2
mkilar123: @mtarking  - indeed, I didn't have the way of testing with 3.1.
I'll try arranging the 3.1 version and re run the tests.

mkilar123: I just tested in 3.1 - the parameter is simply ignored by NDFC.
Even though it's not supported in 3.1 it doesn't cause any conflict and vrf is correctly created/deployed with all other attributes specified.

roles/dtc/common/templates/ndfc_vrfs/msd_fabric/child_fabric/3.2/msd_child_fabric_vrf.j2
juburnet: @mkilar123 and @mtarking - I think this fix also needs to be applied to the msd template: https://github.com/netascode/ansible-dc-vxlan/blob/d6b871d3961bafa32299fdd34a51e7c8e8e76673/roles/dtc/common/templates/ndfc_vrfs/msd_fabric/child_fabric/msd_child_fabric_vrf.j2
The 4.1 MSD pipeline is also failing with the same error as 3.2
```
"For fabric nac-fabric2 and VRF NaC-VRF01; Change of v6VrfRouteMap is not allowed, VRF:NaC-VRF01 is under member fabric and edit is allowed only for fabric instance parameters"
```

mtarking: > @mtarking "v6VrfRouteMap" is implemented exactly the same way as "vrfRouteMap":
> 
> 1. Both have the default value assigned at the dcnm_vrf module level (at base collection).
> 2. At the child fabric-level the value is not editable (it's grey-ed out). It can only be set at the MSD level.
> 3. In the situation when this is skipped at the yaml level, we have the defaults at both SOLUTION and BASE module - so it will get set to FABRIC-RMAP-REDIST-SUBNET - exactly the same way as it happens for "vrfRouteMap"

I do not think this is exactly correct.

When I query, I get back this snippet from the json encoded string:```"v6VrfRouteMap\":\"\"```

mikewiebe: @mkilar123 and @mtarking I verified this in my setup and I see that the default state for `v6VrfRouteMap` is set to `FABRIC-RMAP-REDIST-SUBNET` so I believe we are ok here

roles/dtc/common/templates/ndfc_vrfs/dc_vxlan_fabric/dc_vxlan_fabric_vrfs.j2
mkilar123: version check added to the template

PR #659, opened 2025-10-28
VRF (MSD) Base Module and Action Plugin Support

AKDRG:
1. Fixed and added properties in msd_fabric_vrfs.j2.
2. Fixing VRF changed flags for tasks in vrfs_network.yaml to use the action plugin result.


mikewiebe:
This PR has a dependency on PR https://github.com/CiscoDevNet/ansible-dcnm/pull/530 in the base collection.  The base collection PR should be merged first.

PR #754, opened 2026-03-20
fix: ND 4.2 / NDFC 12.5 compatibility - null interfaces and empty overridden config

juburnet:
## Related Collection Role

* [x] other

## Related Data Model Element

* [ ] vxlan.fabric
* [ ] vxlan.global
* [x] vxlan.topology
* [ ] vxlan.underlay
* [ ] vxlan.overlay
* [ ] vxlan.policy
* [ ] vxlan.multisite
* [ ] defaults.vxlan
* [ ] other

## Proposed Changes

Fix 1 — `prep_106_topology_interfaces.py`: null `interfaces` field

**File:** `plugins/action/common/prepare_plugins/prep_106_topology_interfaces.py`

### Root Cause
In ND 4.2 / NDFC 12.5, the `switchesByFabric` API now returns `null` for the `interfaces` field on switches in **ISN** and **External** fabric types, instead of returning an empty list `[]` as in previous versions.

The prepare plugin iterated directly over this value:
\`\`\`python
Before (crashes when interfaces is null)
for interface in switch.get('interfaces'):
\`\`\`

Iterating over `None` raises `'NoneType' object is not subscriptable`.

Fix
Coerce `null` to an empty list using `or []`:

\`\`\`python
After (safe for null and empty list)
for interface in (switch.get('interfaces') or []):
\`\`\`

---

Fix 2 — `roles/dtc/remove/tasks/common/interfaces.yml`: empty config crashes `cisco.dcnm.dcnm_interface`
**File:** `roles/dtc/remove/tasks/common/interfaces.yml`

### Root Cause
The **Remove Unmanaged Fabric Interfaces** task calls `cisco.dcnm.dcnm_interface` with `state: overridden` and `config: "{{ vars_common_local.interface_all_remove_overridden }}"`. For ISN and External fabrics, this list is always empty (`[]`).

On ND 4.2, the `cisco.dcnm.dcnm_interface` module crashes with `'NoneType' object is not subscriptable` when called with `state: overridden` and an empty `config` list. This is a bug in the upstream `cisco.dcnm` collection against the ND 4.2 API, but can be guarded against on our side.

### Fix
Add a length guard so the task is skipped entirely when there are no interfaces to override:

\`\`\`yaml
Added condition:
- vars_common_local.interface_all_remove_overridden | length > 0
\`\`\`

This prevents `dcnm_interface` from being called with an empty list, which is a no-op anyway.

## Error messages before this fix
TASK [cisco.nac_dc_vxlan.remove : Remove Unmanaged Fabric Interfaces in Nexus Dashboard - Diff Run Feature Disabled] ***
Friday 20 March 2026  18:56:23 +0000 (0:00:00.047)       0:03:21.507 ********** 
Friday 20 March 2026  18:56:23 +0000 (0:00:00.047)       0:03:21.507 ********** 
fatal: [nac-isn1]: FAILED! => {"changed": false, "module_stderr": "'NoneType' object is not subscriptable", "module_stdout": "", "msg": "MODULE FAILURE\nSee stdout/stderr for the exact error"}

## Test Notes

ISN and External fabics deployed successfully in pipeline

**ND Version:** 4.2.1.10
**NDFC Version:** 12.5.0.475

## Checklist

### DIFF
--- a/plugins/action/common/prepare_plugins/prep_106_topology_interfaces.py
+++ b/plugins/action/common/prepare_plugins/prep_106_topology_interfaces.py
@@ -66,7 +66,7 @@ class PreparePlugin:
 
         for switch in data_model.get('vxlan').get('topology').get('switches'):
             # loop through interfaces
-            for interface in switch.get('interfaces'):
+            for interface in (switch.get('interfaces') or []):
                 # loop through interface modes direct and count
                 for interface_mode in self.mode_direct:
                     # if interface mode is a direct match, then increment the count for that mode
--- a/roles/dtc/remove/tasks/common/interfaces.yml
+++ b/roles/dtc/remove/tasks/common/interfaces.yml
@@ -97,6 +97,7 @@
   register: int_data
   when:
     - switch_list.response.DATA | length > 0
+    - vars_common_local.interface_all_remove_overridden | length > 0
     - (interface_delete_mode is defined) and (interface_delete_mode is true|bool)
     - run_map_read_result.diff_run is false|bool or force_run_all is true|bool
 


juburnet:
This fix been running successfully in our pipeline for all ND versions. No issues creating External or ISN fabrics.

dacasti2mx:
# Peer Review — PR #754

## Test Environment
- **ND Version:** 4.2.1.10 / NDFC 12.5.0.475
- **Fabric:** `nac-mcfg-isn` (ISN type, greenfield MCFG deployment)
- **Branch tested:** `fix/nd42-prep106-none-interfaces`
- **Comparison environment:** ND 4.1 (NDFC 12.4) with same ISN fabric definition

## Fixes in this PR — Both Confirmed Working

**Fix 1** (`prep_106_topology_interfaces.py`): Confirmed. The `interfaces: null` coercion to `[]` works correctly. ISN fabric completes `create` and `deploy` roles successfully (191 ok, 27 changed).

**Fix 2** (`interfaces.yml` guard): Confirmed. The `length > 0` guard correctly skips `dcnm_interface` when `interface_all_remove_overridden` is empty.

## Additional Failure — Not Covered by This PR

When `interface_all_remove_overridden` is **non-empty** (our ISN has a loopback `lo0`), the guard passes and `dcnm_interface` is called with `state: overridden`. The module then crashes:

```
TASK [cisco.nac_dc_vxlan.remove : Remove Unmanaged Fabric Interfaces - Diff Run Feature Disabled]
fatal: [nac-mcfg-isn]: FAILED! => {
    "module_stderr": "'NoneType' object is not subscriptable"
}
```

### Exact crash location

We obtained the full traceback by running with `ANSIBLE_NETWORK_IMPORT_MODULES=0`:

```
File ".../cisco/dcnm/plugins/modules/dcnm_interface.py", line 5338, in dcnm_intf_get_diff_overridden
TypeError: 'NoneType' object is not subscriptable
```

The failing line in `dcnm_interface.py`:

```python
# Line 5331-5338
for intf in defer_list:
    delem = {}
    sno = intf["serialNo"]
    fabric = intf["fabricName"]
    name = intf["underlayPolicies"][0]["source"]   # <-- crash here
```

`underlayPolicies` is `null` for these interfaces, so `None[0]` raises `TypeError`.

### Why interfaces end up in `defer_list`

The `dcnm_intf_get_diff_overridden` function queries NDFC for all interfaces on the switch. Interfaces with `deletable: false` are placed in `defer_list` (line 5045-5047) to be processed later. The code at line 5338 then assumes every deferred interface has a populated `underlayPolicies` list, which is not the case.

### Root cause — API behavior change between NDFC versions

We compared the same API call across versions:

```
GET /appcenter/cisco/ndfc/api/v1/lan-fabric/rest/interface/detail?serialNumber=9A7R7ZDGB6U
```

(Serial number `9A7R7ZDGB6U` = `isn-rtr1`)

| Field | ND 4.1 ISN | ND 4.2 ISN |
|-------|-----------|-----------|
| Interfaces with `underlayPolicies: null` | 64 | 65 |
| Interfaces with `deletable: false` | **0** | **65** |

Both versions return `underlayPolicies: null` for Ethernet interfaces on ISN fabrics. The difference is that **NDFC 12.5 (ND 4.2) changed `deletable` from `true` to `false`** for these interfaces.

In ND 4.1, those Ethernets have `deletable: true`, so they never enter `defer_list` — the code at line 5338 is never reached. In ND 4.2, they have `deletable: false`, enter `defer_list`, and the code crashes when accessing `underlayPolicies[0]` on a `null` value.

We also confirmed that an External fabric on ND 4.1 shows `underlayPolicies: null` on most Ethernets but with `deletable: false` only on 2 interfaces that have `underlayPolicies` populated — consistent with the pattern that previous NDFC versions never put interfaces with `underlayPolicies: null` into the `defer_list`.

### Impact

- **Create + deploy succeed** — the ISN fabric is fully functional
- The failure occurs only in the final `remove` reconciliation phase
- This is not a regression in greenfield deployment

### Recommendation

1. **This PR should be merged as-is** — both fixes solve real problems
2. **Open a companion issue on [CiscoDevNet/ansible-dcnm](https://github.com/CiscoDevNet/ansible-dcnm)** for the upstream crash in `dcnm_interface.py` line 5338, with the evidence above
3. **Optional workaround for this PR**: add a fabric-type guard to skip `dcnm_interface` override entirely for ISN/External fabrics where Ethernet interface override is generally not applicable:
   ```yaml
   - data_model_extended.vxlan.fabric.type not in ["ISN", "External"]
   ```

dacasti2mx (approved):
LGTM tested with 4.1 and 4.2

dacasti2mx:
opened bug in the dcnm collection repo.
https://github.com/CiscoDevNet/ansible-dcnm/issues/658

juburnet:
Reproduced the issue on a call with @mikewiebe. This fix will not be committed, but instead fixed in DCNM collection.

PR #773, opened 2026-04-15
Diff underlay ip address

ccoueffe:
## Related Collection Role

* [ ] other

## Related Data Model Element

* [ ] vxlan.fabric
* [ ] vxlan.global
* [ ] vxlan.topology
* [ ] vxlan.underlay
* [ ] vxlan.overlay
* [ ] vxlan.policy
* [ ] vxlan.multisite
* [ ] defaults.vxlan
* [ ] other

## Test Notes

Without change:

```
────────────────────────────────────────────────────────────────────────
CREATE [nac-fabric1] Step 3/20: underlay_ip_address (_underlay_ip_remote_diff)
────────────────────────────────────────────────────────────────────────
underlay_ip_filter [nac-fabric1]: fetched 88 items from controller, pools=['anycast_rp_ip_pool', 'loopback0_ip_pool', 'loopback1_ip_pool']
underlay_ip_filter [nac-fabric1]: 11 items after pool/scope filter
underlay_ip_filter [nac-fabric1]: 11 matched, 0 need update
CREATE [nac-fabric1] underlay_ip_address: 11 configured → 0 after audit
CREATE [nac-fabric1] underlay_ip_address: 11 matched controller, 0 need update
CREATE [nac-fabric1] underlay_ip_address → ok (changed=False) [0.6s]

─────────────────────────────────────────────────────────────────────────
CREATE [nac-fabric1] Step 4/20: underlay_ip_address (dcnm_resource_manager)
─────────────────────────────────────────────────────────────────────────
CREATE [nac-fabric1] underlay_ip_address → skipped (no diff) [0.0s]
```

With one change in the data model:
```
─────────────────────────────────────────────────────────────────────────
CREATE [nac-fabric1] Step 3/20: underlay_ip_address (_underlay_ip_remote_diff)
─────────────────────────────────────────────────────────────────────────
underlay_ip_filter [nac-fabric1]: fetched 88 items from controller, pools=['loopback0_ip_pool', 'loopback1_ip_pool', 'anycast_rp_ip_pool']
underlay_ip_filter [nac-fabric1]: 11 items after pool/scope filter
underlay_ip_filter [nac-fabric1]: 10 matched, 1 need update
  [nac-fabric1] resource_mismatch: entity=9SLMV1LE2LT~loopback0 expected=100.0.1.155 actual=100.0.0.155
CREATE [nac-fabric1] underlay_ip_address: 11 configured → 1 after audit
CREATE [nac-fabric1] underlay_ip_address: 10 matched controller, 1 need update
  [nac-fabric1] resource_mismatch: entity=9SLMV1LE2LT~loopback0 expected=100.0.1.155 actual=100.0.0.155
CREATE [nac-fabric1] underlay_ip_address → ok (changed=False) [0.6s]

─────────────────────────────────────────────────────────────────────────
CREATE [nac-fabric1] Step 4/20: underlay_ip_address (dcnm_resource_manager)
─────────────────────────────────────────────────────────────────────────
CREATE [nac-fabric1] Executing dcnm_resource_manager for underlay_ip_address with state=merged, save=None, deploy=None, items=1
CREATE [nac-fabric1] underlay_ip_address → ok (changed=True) [2.5s]


<!-- Please provide Cisco Nexus Dashboard version developed against -->


## Checklist

### DIFF
--- a/plugins/action/dtc/manage_resources.py
+++ b/plugins/action/dtc/manage_resources.py
@@ -206,17 +206,38 @@ class ResourceManager(PipelineRunnerBase):
 
         return {'changed': False}
       
-    def _underlay_ip_query_and_filter(self, resource_name, step):
+    def _underlay_ip_remote_diff(self, resource_name, step):
         """
-        Audit underlay IP allocations against existing ND pool resources.
+        Remote diff: reconcile underlay IP data model against controller state.
 
-        Mirrors the old roles/dtc/create/tasks/common/underlay_ip_address.yml flow:
-          1. Resolve underlay_ip_address data (respecting diff_run)
-          2. For full runs, audit desired config against ND pool resources
-          3. Replace module_data with filtered_config from the audit plugin
+        This method implements Phase 2 of the two-phase comparison:
 
-        In diff_run mode, preserve the existing diff.updated narrowing and skip
-        the expensive controller audit, matching the old behavior.
+        Phase 1 - Local YAML Comparison (handled by common/files):
+            Compares the previous and current rendered YAML
+            (ndfc_underlay_ip_address.yml) to detect local data model changes.
+            Output: diff.updated entries for changed resources.
+
+            Example: user changes loopback0 ipv4 from 10.1.0.1/32
+            to 10.1.0.2/32 in the data model -> Phase 1 detects the diff.
+
+        Phase 2 - Controller Reconciliation (this method):
+            Queries the controller (ND) pool resources via
+            underlay_ip_manual_allocation_filter and compares against the desired config.
+            Reduces module_data to only entries needing controller updates.
+
+            Example: 5 switches in data model, but 2 already have the
+            correct pool allocation on the controller -> filtered_config
+            returns only the 3 that need updates.
+
+        In diff_run mode the local YAML diff (Phase 1) already narrows scope,
+        so Phase 2 is skipped to avoid redundant controller queries.
+
+        Args:
+            resource_name: Resource identifier (e.g. underlay_ip_address).
+            step: Pipeline step dict from create_resources.yml.
+
+        Returns:
+            dict with changed and optionally failed/msg keys.
         """
         data, _ = self._resolve_step_data(resource_name, step)
         if not data:
@@ -236,14 +257,14 @@ class ResourceManager(PipelineRunnerBase):
             module_args['query_pools'] = self.task_vars['underlay_ip_audit_pools']
 
         audit_result = self.executor.execute_plugin(
-            module_name='cisco.nac_dc_vxlan.dtc.underlay_ip_pool_audit',
+            module_name='cisco.nac_dc_vxlan.dtc.underlay_ip_manual_allocation_filter',
             module_args=module_args,
         )
 
         if audit_result.get('failed'):
             return {
                 'failed': True,
-                'msg': audit_result.get('msg', 'underlay_ip_pool_audit failed'),
+                'msg': audit_result.get('msg', 'underlay_ip_manual_allocation_filter failed'),
             }
 
         filtered_config = audit_result.get('filtered_config', [])
--- a/plugins/action/dtc/manage_resources.py
+++ b/plugins/action/dtc/manage_resources.py
@@ -205,7 +205,61 @@ class ResourceManager(PipelineRunnerBase):
         )
 
         return {'changed': False}
+      
+    def _underlay_ip_query_and_filter(self, resource_name, step):
+        """
+        Audit underlay IP allocations against existing ND pool resources.
 
+        Mirrors the old roles/dtc/create/tasks/common/underlay_ip_address.yml flow:
+          1. Resolve underlay_ip_address data (respecting diff_run)
+          2. For full runs, audit desired config against ND pool resources
+          3. Replace module_data with filtered_config from the audit plugin
+
+        In diff_run mode, preserve the existing diff.updated narrowing and skip
+        the expensive controller audit, matching the old behavior.
+        """
+        data, _ = self._resolve_step_data(resource_name, step)
+        if not data:
+            return {'changed': False, 'msg': 'No underlay_ip_address data to audit'}
+
+        if self.run_map_diff_run:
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: plugins/action/dtc/underlay_ip_manual_allocation_filter.py [hunk 1]
```diff
@@ -0,0 +1,476 @@
...
+            }
+            pools.discard("")
+
+        # === BULK QUERY: SINGLE REST CALL FOR ALL FABRIC RESOURCES ===
+        # Fetch all allocations in one call rather than per-pool queries to minimize
+        # controller load and round-trip latency. Filtering happens client-side.
+        path = "/appcenter/cisco/ndfc/api/v1/lan-fabric/rest/resource-manager/fabrics/{}".format(
+            fabric
+        )
+        all_result = self._execute_ndfc_rest("GET", path, task_vars, tmp)
```
mikewiebe, line 255: @ccoueffe  Did some analysis on the client code that processes all of these allocations that come back from the GET call in terms of performance

The PR code is roughly linear, not obviously quadratic: O(controller_allocations + desired_config). The issue is memory and repeated passes:

* _extract_resource_items() recursively builds a full flattened all_existing_items list and copies each item with dict(data).
* Then it builds filtered_existing_items, a second list.
* Then it loops again to build existing_map, existing_link_map, and existing_pool_resource_set.
* Then it loops over desired_config.
* At verbose level 2, it also logs list(pools), which could be large if many subnet pools are auto-detected.

So if /resource-manager/fabrics/{fabric} returns thousands or tens of thousands of allocations, the code pays for the full REST payload, full JSON parse, full flatten, then multiple in-memory structures before it ever compares desired state.

A better implementation would either query only relevant pools/scopes if the API supports it, or make _extract_resource_items() a generator and combine extraction, pool/scope filtering, and index-building in one pass. That would avoid all_existing_items and filtered_existing_items as separate materialized lists.

Not sure how many resources we are managing or will grow to manage here?

plugins/action/dtc/underlay_ip_manual_allocation_filter.py
```diff
@@ -0,0 +1,476 @@
...
+
+        query_errors = []
+        all_existing_items = []
+        if all_result.get("failed"):
+            query_errors.append({"path": path, "msg": all_result.get("msg")})
+        else:
+            response = all_result.get("response", {})
+            all_existing_items = _extract_resource_items(response)
+
+        display.vv(
```
mikewiebe: pools is auto-built from every desired pool_name. For fabric links, endpoint allocations use subnet values as pool names, so large fabrics can create a very large pool set. Membership is still O(1), but it increases memory and makes display.vv(... pools={list(pools)}) noisy and expensive at verbosity 2.

I think I am ok with this since we only see it using -vv but just want to make sure this is consistent with other uses of -vv in our collection

plugins/action/dtc/underlay_ip_manual_allocation_filter.py
```diff
@@ -0,0 +1,476 @@
...
+
+            # Tier 2: Canonical link match for bidirectional links
+            if existing is None and normalized_scope == "link":
+                existing = existing_link_map.get(
+                    (_canonicalize_link_entity(entity_name), normalized_pool)
+                )
+                if existing is not None:
+                    matched_by = "canonical_link"
+
+            # Tier 3: Pool+resource fallback (resource already allocated, entity may differ)
```
mikewiebe: The pool/resource fallback treats an allocation as matched even when the controller has the same resource assigned to a different entity. Because existing_pool_resource_set is populated only after entity_name exists, this fallback does not help entity-less responses; it only hides ownership mismatches. 

Example: 

desired says `serialA~loopback1` should own `10.1.1.1/32`, but NDFC has that resource under `serialB~loopback1`. The filter increments matched_count, omits the item from filtered_config, and the later dcnm_resource_manager step never gets a chance to repair or fail visibly. 

Can you test this scenario @ccoueffe to see if it's a problem?

Suggestion from AI  is to remove this fallback or make it explicitly conditional on a response shape that truly lacks entity identity.

plugins/action/dtc/manage_resources.py
```diff
@@ -205,7 +205,100 @@ def _fabric_links_query_and_filter(self, resource_name, step):
...
+
+        scope_filter = self.task_vars.get("underlay_ip_audit_scope_filter", "all")
+
+        module_args = {
+            "fabric": self.fabric_name,
+            "desired_config": data,
+            "scope_filter": scope_filter,
+        }
+        if "underlay_ip_audit_pools" in self.task_vars:
+            module_args["query_pools"] = self.task_vars["underlay_ip_audit_pools"]
```
mtarking: Where is underlay_ip_audit_pools getting set initially?

plugins/action/dtc/underlay_ip_manual_allocation_filter.py
ccoueffe: Merged extract/filter/index into a single-pass loop. Eliminated intermediate lists.

plugins/action/dtc/manage_resources.py
ccoueffe: Good catch — not set anywhere, it was dead code. Removed both underlay_ip_audit_pools and underlay_ip_audit_scope_filter from manage_resources.py. The filter now auto-detects pools from desired_config.

plugins/action/dtc/underlay_ip_manual_allocation_filter.py
ccoueffe: Removed Tier 3 entirely. Two-tier matching only: exact entity+pool match, then canonical link fallback.

plugins/action/dtc/underlay_ip_manual_allocation_filter.py
ccoueffe: Moved pools list to display.vvv. Item counts remain at display.v/display.vv.

ccoueffe:
Note: after dcnm_resource_manager updates pool allocations, a **Recalculate & Deploy** (save) is required for changes to be applied to device configuration.

mikewiebe (approved):

mtarking (approved):

PR #775, opened 2026-04-17
Add 'env_var_' prefix option for handling secrets

### LINKED ISSUE #720: Support for 'env_var_' prefix option to secure any data model secret

cisco.nac_dc_vxlan.dtc.create

### New feature description:

My customer requires to not store any secrets in the YAML configuration files. He would like to refer to environment variables defined in his CI/CD system.

Currently NaC VXLAN supports environment variable lookup using the 'env_var_' prefix specifically for switch credentials. I propose to extend this option to support any data model secret. Example:

```YAML
---
vxlan:
  underlay:
    bgp:
      authentication_enable: true
      authentication_key_type: 3
      authentication_key: env_var_bgp_auth_key
```

```shell
n/a
```

vxlan.underlay

```shell
n/a
```

Need to extend credentials guide.

```yaml
n/a
```

### Ansible Version

```shell
ansible [core 2.16.3]
```

```shell
Collection         Version
------------------ -------
ansible.netcommon  7.1.0  
ansible.posix      2.0.0  
ansible.utils      5.1.2  
cisco.dcnm         3.10.0 
cisco.nac_dc_vxlan 0.6.0  
cisco.nxos         9.4.0  
community.general  10.1.0
```

```shell
3.2(1i)
```

### Cisco NX-OS Version

```shell
10.5(3)
```

marehler:
## Related Issue(s)

Fixes #720 

## Related Collection Role

* [ ] other

## Related Data Model Element

* [ ] vxlan.fabric
* [x] vxlan.global
* [ ] vxlan.topology
* [x] vxlan.underlay
* [ ] vxlan.overlay
* [x] vxlan.policy
* [x] vxlan.multisite
* [ ] defaults.vxlan
* [ ] other

## Proposed Changes

Add plugin to lookup up secrets starting with the 'env_var_' prefix. Same functionality as already supported for switch credentials. Examples:
```
---
---
vxlan:
  underlay:
    bgp:
      max_paths: 4
      authentication_enable: true
      authentication_key_type: 3
      authentication_key: env_var_bgp_auth_key
    multicast:
      ipv4:
        authentication_enable: true
        authentication_key: env_var_mcast_auth_key
  global:
    ebgp:
      aaa_freeform: |
        feature tacacs+
        tacacs-server key 7 env_var_tacacs_key
        ip tacacs source-interface mgmt0
        tacacs-server timeout 20
```
- Created plugins/action/common/prepare_plugins/prep_005_resolve_env_vars.py
- Runs as prepare plugin 005 in the prepare_service_model pipeline — after fabric/global setup and list defaults (001–004), before any topology processing (104+)
- Recursively walks the entire model_extended data model
- Any string value starting with env_var_ is resolved via os.getenv()
- If the env var isn't set, a warning is emitted and the original string is kept unchanged
- Verbose logging (-v) shows how many variables were resolved; -vvv shows each individually with its data model path
- No impact on existing switch credential flow: get_credentials.py operates on the rendered inv_list (after templates), not the raw data model. The two mechanisms are complementary and don't interfere.

## Test Notes

1) BGP Fabric
```
TASK [cisco.nac_dc_vxlan.validate : Prepare Service Model] ***************************************************************************************************
task path: /Users/marehler/Git/sac-ndfc/nac-ndfc/collections/ansible_collections/cisco/nac_dc_vxlan/roles/validate/tasks/sub_main.yml:110
Resolved 'env_var_mcast_auth_key' from environment variable at 'vxlan.underlay.multicast.ipv4.authentication_key'
Resolved 'env_var_bgp_auth_key' from environment variable at 'vxlan.underlay.bgp.authentication_key'
Resolved 'env_var_tacacs_key' from environment variable at 'vxlan.global.ebgp.aaa_freeform'
Resolved 3 environment variable(s) in the data model
```
2) VXLAN EVPN Fabric
```
TASK [cisco.nac_dc_vxlan.validate : Prepare Service Model] *******************************************************************
task path: /Users/marehler/Git/sac-ndfc/nac-ndfc/collections/ansible_collections/cisco/nac_dc_vxlan/roles/validate/tasks/sub_main.yml:110
Resolved 'env_var_ospf_auth_key' from environment variable at 'vxlan.underlay.ospf.authentication_key'
Resolved 1 environment variable(s) in the data model
```
3) MSD Fabric
```
TASK [cisco.nac_dc_vxlan.validate : Prepare Service Model] ***************************************************************************************************
task path: /Users/marehler/Git/sac-ndfc/nac-ndfc/collections/ansible_collections/cisco/nac_dc_vxlan/roles/validate/tasks/sub_main.yml:110
Resolved 'env_var_dci_password' from environment variable at 'vxlan.multisite.overlay_dci.ebgp_password'
Resolved 1 environment variable(s) in the data model
````

3.2.2
4.2.1
4.3.1

## Checklist

### DIFF
--- a/docs/Handling_Secrets_Guide.md
+++ b/docs/Handling_Secrets_Guide.md
@@ -1,0 +1,136 @@
+# Handling Secrets Guide
+
+This guide describes how to manage secrets such as passwords and authentication keys in the Network as Code VXLAN data model. There are two mechanisms available:
+
+1. **Ansible Vault** - Encrypts entire variable values at rest.
+2. **Environment Variable Lookup (`env_var_`)** - Resolves secrets from environment variables at runtime. Suitable for CI/CD integration.
+
+Both approaches ensure that plaintext secrets are not visible in the data YAML files. The key difference is that Ansible Vault encrypts whole values while `env_var_` replaces individual strings — making `env_var_` the only option for secrets embedded in freeform configuration blocks.
+
+---
+
+## Ansible Vault
+
+[Ansible Vault](https://docs.ansible.com/ansible/latest/vault_guide/index.html) encrypts sensitive data so it can be stored safely in data files. Encrypted values are decrypted automatically at playbook runtime when a vault password is provided.
+
+### Encrypting a Value
+
+```bash
+ansible-vault encrypt_string 'MySecretPassword' --name 'authentication_key'
+```
+
+This produces an encrypted variable which you can paste into your data files:
+
+```yaml
+vxlan:
+  underlay:
+    bgp:
+      authentication_enable: true
+      authentication_key_type: 3
+      authentication_key: !vault |
+        $ANSIBLE_VAULT;1.1;AES256
+        61626364656667686970...
+```
+
+### Running Playbooks with Vault
+
+```bash
+# Prompt for vault password
+ansible-playbook -i inventory deploy.yaml --ask-vault-pass
+
+# Use a password file
+ansible-playbook -i inventory deploy.yaml --vault-password-file ~/.vault_pass
+```
+
+### Limitation: Freeform Configurations
+
+Ansible Vault encrypts the entire value of a YAML field. This works for standalone fields like `authentication_key` where the complete value is a secret.
+
+However, freeform configuration blocks may contain a mix of regular configuration and secrets within a single multi-line string:
+
+```yaml
+# Does not work with Ansible Vault because only the TACACS key is a secret,
+# not the entire block.
+aaa_freeform: |
+  feature tacacs+
+  tacacs-server key 7 MySecretTacacsKey
+  ip tacacs source-interface mgmt0
+  tacacs-server timeout 20
+```
+
+Encrypting the entire `aaa_freeform` value with Ansible Vault would encrypt all four lines, which is not desired.
+
+---
+
+## Environment Variable Lookup (`env_var_`)
+
+The `env_var_` prefix instructs NaC to resolve a value from an environment variable at runtime. This works for:
+
+- **Standalone values** - the entire YAML value is an reference to an environment variable
+- **Embedded secrets** - `env_var_` references inside larger strings (e.g. freeform configuration blocks)
+
+### How It Works
+
+Any string value in the data model that contains `env_var_` is resolved by the NaC VXLAN `validate` role. The secret is replaced with the value of the corresponding environment variable.
+
+* Variable names in YAML must exactly match the environment variable names (including the env_var_ prefix).
+* Valid environment variable names may include letters, digits, and underscores.
+* Use single quotes to prevent shell interpretation of special characters
+* If the environment variable is not set, a warning is displayed and the secret is left unchanged.
+
+### Setting Environment Variables
+
+```bash
+export env_var_BGP_AUTH_KEY='MyBgpP@$$w0rd'
+export env_var_MCAST_AUTH_KEY='mcast_secret_123'
+export env_var_TACACS_KEY='TacacsSecret!'
+export env_var_DCI_PASSWORD='DciP@ssword'
+```
+
+### Examples: Standalone Values
+
+```yaml
+vxlan:
+  underlay:
+    multicast:
+      ipv4:
+        authentication_enable: true
+        authentication_key: env_var_MCAST_AUTH_KEY
+    bgp:
+      authentication_enable: true
+      authentication_key_type: 3
+      authentication_key: env_var_BGP_AUTH_KEY
+```
+
+```yaml
+vxlan:
+  multisite:
+    overlay_dci:
+      enable_ebgp_password: True
+      ebgp_password: env_var_DCI_PASSWORD
+      ebgp_password_encryption_type: 3
+```
+
+### Example: Freeform Configurations
+
+```yaml
+vxlan:
+  global:
+    ebgp:
+      aaa_freeform: |
+        feature tacacs+
+        tacacs-server key 7 env_var_TACACS_KEY
+        ip tacacs source-interface mgmt0
+        tacacs-server timeout 20
+```
+
+At runtime, only `env_var_TACACS_KEY` is replaced with the value of the `env_var_TACACS_KEY` environment variable. The rest of the block is unchanged:
+
+```
+feature tacacs+
+tacacs-server key 7 TacacsSecret!
+ip tacacs source-interface mgmt0
+tacacs-server timeout 20
+```
+
+This pattern works in any freeform field (`aaa_freeform`, `banner_freeform`, `bootstrap_freeform`, `intra_fabric_link_freeform`, `freeform_config`, etc.) and supports multiple `env_var_` tokens in the same block.
--- a/docs/Handling_Secrets_Guide.md
+++ b/docs/Handling_Secrets_Guide.md
@@ -86,7 +86,6 @@ Any string value in the data model that contains `env_var_` is resolved by NaC V
 The value stored in each variable must already be encoded for the field that references it. The placeholders below stand for the encrypted (ciphertext) values, not plaintext secrets:
 
 ```bash
-# Each value below is the ciphertext for its field's encryption type, not a plaintext secret.
 export env_var_BGP_AUTH_KEY='<3DES-encrypted BGP authentication key>'
 export env_var_MCAST_AUTH_KEY='<3DES-encrypted PIM hello key>'
 export env_var_TACACS_KEY='<type-7 encoded TACACS key>'
--- a/docs/Handling_Secrets_Guide.md
+++ b/docs/Handling_Secrets_Guide.md
@@ -145,3 +145,5 @@ This pattern works in any freeform field (`aaa_freeform`, `banner_freeform`, `bo
 * The `env_var_` lookup mechanism is not supported for policies referencing a file.
 
 * If an environment variable changes, the change will not be detected in case of a diff run (enabled by default), because the configuration in the data YAML files still looks the same. To enforce a new lookup, you can set the `force_run_all` parameter to true.
+
+* Secrets are not shown in rendered payload files, but the resolved value is handled in memory and may therefore still be visible in logs. This applies to NaC collection tasks as well as the `cisco.dcnm` collection. When `cisco.dcnm` logging is enabled, its modules can log the request parameters, payload, and controller response, which include the resolved value.
--- a/docs/Handling_Secrets_Guide.md
+++ b/docs/Handling_Secrets_Guide.md
@@ -71,20 +71,26 @@ The `env_var_` prefix instructs NaC to resolve a value from an environment varia
 
 ### How It Works
 
-Any string value in the data model that contains `env_var_` is resolved by NaC VXLAN. The secret is replaced with the value of the corresponding environment variable.
+Any string value in the data model that contains `env_var_` is resolved by NaC VXLAN. The string is replaced with the value of the corresponding environment variable.
 
+* Resolution is global. Any string in the data model that contains `env_var_` followed by word characters (`env_var_\w+`) is treated as a token, not only known secret fields. This allows `env_var_` to work inside freeform blocks, but it also means ordinary text that happens to contain the prefix is rewritten.
 * Variable names in YAML must exactly match the environment variable names (including the env_var_ prefix).
 * Valid environment variable names may include letters, digits, and underscores.
-* Use single quotes to prevent shell interpretation of special characters
-* If a referenced environment variable is not set, the run fails during the `validate` stage. The error lists every missing variable with its location in the data model. This behavior prevents an unresolved `env_var_` token from being sent to Nexus Dashboard as a literal credential value.
+* If a referenced environment variable is not set, the run fails during the `validate` stage. The error lists every missing variable with its location in the data model. This behavior prevents an unresolved `env_var_` string from being sent to Nexus Dashboard as a literal credential value.
+* `env_var_` performs substitution only, it does not encrypt. The resolved value is passed to Nexus Dashboard exactly as stored in the environment variable, with no transformation. It is the operator's responsibility to store a value that already matches what the target field expects.
+* For fields that carry an encryption type (for example `authentication_key` with `authentication_key_type: 3`, `ebgp_password` with `ebgp_password_encryption_type: 3`, or a `tacacs-server key 7 ...` line in a freeform block), the environment variable must contain the value already encoded for that type (the ciphertext), not the plaintext secret.
+* To keep literal `env_var_...` text (for example in a banner) from being resolved, avoid the exact `env_var_` prefix. For example break the word so it does not match, or reword the text. There is no dedicated escape sequence.
 
 ### Setting Environment Variables
 
+The value stored in each variable must already be encoded for the field that references it. The placeholders below stand for the encrypted (ciphertext) values, not plaintext secrets:
+
 ```bash
-export env_var_BGP_AUTH_KEY='MyBgpP@$$w0rd'
-export env_var_MCAST_AUTH_KEY='mcast_secret_123'
-export env_var_TACACS_KEY='TacacsSecret!'
-export env_var_DCI_PASSWORD='DciP@ssword'
+# Each value below is the ciphertext for its field's encryption type, not a plaintext secret.
+export env_var_BGP_AUTH_KEY='<3DES-encrypted BGP authentication key>'
+export env_var_MCAST_AUTH_KEY='<3DES-encrypted PIM hello key>'
+export env_var_TACACS_KEY='<type-7 encoded TACACS key>'
+export env_var_DCI_PASSWORD='<3DES-encrypted DCI password>'
 ```
 
 ### Examples: Standalone Values
@@ -124,11 +130,11 @@ vxlan:
         tacacs-server timeout 20
 ```
 
-At runtime, only `env_var_TACACS_KEY` is replaced with the value of the `env_var_TACACS_KEY` environment variable. The rest of the block is unchanged:
+At runtime, only `env_var_TACACS_KEY` is replaced with the value of the `env_var_TACACS_KEY` environment variable. Because the line uses `key 7`, the variable must contain the type-7 encoded key (not the plaintext). The rest of the block is unchanged:
 
 ```
 feature tacacs+
-tacacs-server key 7 TacacsSecret!
+tacacs-server key 7 076a0f1e1d0a19173e243f36
 ip tacacs source-interface mgmt0
 tacacs-server timeout 20
 ```
--- a/docs/Handling_Secrets_Guide.md
+++ b/docs/Handling_Secrets_Guide.md
@@ -76,7 +76,7 @@ Any string value in the data model that contains `env_var_` is resolved by NaC V
 * Variable names in YAML must exactly match the environment variable names (including the env_var_ prefix).
 * Valid environment variable names may include letters, digits, and underscores.
 * Use single quotes to prevent shell interpretation of special characters
-* If the environment variable is not set, a warning is displayed and the secret is left unchanged.
+* If a referenced environment variable is not set, the run fails during the `validate` stage. The error lists every missing variable with its location in the data model. This behavior prevents an unresolved `env_var_` token from being sent to Nexus Dashboard as a literal credential value.
 
 ### Setting Environment Variables
 
--- a/docs/Handling_Secrets_Guide.md
+++ b/docs/Handling_Secrets_Guide.md
@@ -71,7 +71,7 @@ The `env_var_` prefix instructs NaC to resolve a value from an environment varia
 
 ### How It Works
 
-Any string value in the data model that contains `env_var_` is resolved by the NaC VXLAN `validate` role. The secret is replaced with the value of the corresponding environment variable.
+Any string value in the data model that contains `env_var_` is resolved by NaC VXLAN. The secret is replaced with the value of the corresponding environment variable.
 
 * Variable names in YAML must exactly match the environment variable names (including the env_var_ prefix).
 * Valid environment variable names may include letters, digits, and underscores.
@@ -134,3 +134,9 @@ tacacs-server timeout 20
 ```
 
 This pattern works in any freeform field (`aaa_freeform`, `banner_freeform`, `bootstrap_freeform`, `intra_fabric_link_freeform`, `freeform_config`, etc.) and supports multiple `env_var_` tokens in the same block.
+
+### Limitations
+
+* The `env_var_` lookup mechanism is not supported for policies referencing a file.
+
+* If an environment variable changes, the change will not be detected in case of a diff run (enabled by default), because the configuration in the data YAML files still looks the same. To enforce a new lookup, you can set the `force_run_all` parameter to true.
--- a/plugins/action/common/prepare_plugins/prep_005_resolve_env_vars.py
+++ b/plugins/action/common/prepare_plugins/prep_005_resolve_env_vars.py
@@ -1,0 +1,106 @@
+# Copyright (c) 2025 Cisco Systems, Inc. and its affiliates
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy of
+# this software and associated documentation files (the "Software"), to deal in
+# the Software without restriction, including without limitation the rights to
+# use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+# the Software, and to permit persons to whom the Software is furnished to do so,
+# subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

**Comment by marehler** [comment]
Sync'd fork with release 0.8.0.

**Review by skaszlik — CHANGES_REQUESTED** [review]
With this new method of handling the passwords, please investigate how we can merge the `prep_005_resolve_env_vars` plugin with existing `get_credentials.py` to simplify the code. It looks like the `prep_005` can completely replace the `get_credentials.py`

**Comment by marehler** [comment]
Changed implementation so that resolved environment variables are not shown in the rendered files. Summary:

**prep_005_resolve_env_vars.py**
- Before: prepare() called resolve_env_vars_recursive() which replaced env_var_XXX tokens with actual secret values in model_extended → secrets ended up in rendered files
- After: prepare() calls validate_env_vars_recursive() which only checks env vars exist and warns if missing — tokens stay as-is in model_extended
- resolve_env_vars_recursive() is kept and now exported for use by build_resource_data.py

**build_resource_data.py**
- Imports copy and resolve_env_vars_recursive from prep_005
- In build_resource(), after post-hook module_data adoption (line 336): deep-copies module_data, resolves env_var_ tokens in the copy, and uses the resolved copy as module_data
- The rendered file keeps the env_var_XXX placeholders
- The module_data sent to NDFC modules has actual values

Please note env_var_ secrets are still visible in Ansible logs. For switch credentials (get_credentials.py) using the dcnm_inventory module, the module's argument spec marks password and user_name with no_log: True. Ansible automatically redacts any parameter marked this way in task output → VALUE_SPECIFIED_IN_NO_LOG_PARAMETER. However for secrets using env_var_ and the dcnm_fabric module, the module accepts a generic config dict. Individual keys like protocol authentication keys are not marked no_log in the argument spec.

**Comment by marehler** [comment]
> With this new method of handling the passwords, please investigate how we can merge the `prep_005_resolve_env_vars` plugin with existing `get_credentials.py` to simplify the code. It looks like the `prep_005` can completely replace the `get_credentials.py`

`get_credentials.py` has a different scope and fallback behavior. `get_credentials` has smart cascading fallbacks — if an env_var_ credential isn't set, it falls back to group_vars (ndfc_switch_username/ndfc_switch_password). The generic resolver in build_resource_data has no fallback — if the env var isn't set, the token stays as-is and a warning is emitted.
`get_credentials` also does more than env var resolution — it handles per-switch credential lookup from the data model, POAP discovery credentials, and deep-copy credential enrichment. It's a credential workflow plugin that happens to support env_var_ tokens, not a generic env var resolver. In conclusion, I think `prep_005_resolve_env_vars` cannot completely replace it.

**Review by ccoueffe — APPROVED** [review]
LGTM but we need to document in netascode because It doesn't work during validate and we need a force_run_all in that case.

**Comment by marehler** [comment]
- Fixed env var resolution for policies in case of diff run. Root cause: diff.updated is computed from the rendered YAML before `resolve_env_vars_recursive` runs in build_resource_data.py. When `_resolve_create_data` prefers diff.updated during diff_run mode, unresolved env_var_ tokens are sent to ND. Change in manage_resources.py: In `_resolve_create_data`, when taking the diff.updated path, deep-copy the data and run `resolve_env_vars_recursive` on it before returning. Two imports added (copy, resolve_env_vars_recursive).
- Documented limitations as mentioned by Charly.

**Review by skaszlik — APPROVED** [review]
LGTM

**Comment by mikewiebe** [comment]
@marehler Given that we are recursively walking the entire model_extended data model I am concerned about the performance impact at scale.  Have you scaled the data model (above 10MB) with a large number of env_vars* to profile the performance impact?

File: plugins/action/common/prepare_plugins/prep_005_resolve_env_vars.py [hunk 1]
```diff
@@ -0,0 +1,159 @@
...
+    """
+    Resolve a single env_var_ token to its environment variable value.
+
+    Note: Environment variables containing special characters like $, `, \\, etc.
+    should be properly escaped when setting them in the shell.
+    Example: export BGP_AUTH_KEY='MyP@$$w0rd' (use single quotes to prevent shell interpretation)
+    """
+    resolved = os.getenv(token)
+
+    if resolved is None:
```
mikewiebe, line 58: ### High: Missing environment variables are sent to NDFC as literal credential values

**Issue**

The resolver treats a missing environment variable as a warning-only condition and returns the original `env_var_*` token. The active create path then passes that literal string to the base collection as the configured secret.

**Evidence**

- [`resolve_env_var_token()` returns the token when `os.getenv()` is `None`](https://github.com/netascode/ansible-dc-vxlan/pull/775/files#diff-2ddb7c7201fd10a6475b8da4fcef145d41a74c605d6974e450afb68a089968ceR44), and validation similarly only warns at [lines 100–106](https://github.com/netascode/ansible-dc-vxlan/pull/775/files#diff-2ddb7c7201fd10a6475b8da4fcef145d41a74c605d6974e450afb68a089968ceR100).
- The guide explicitly documents that an unset variable is left unchanged at [line 79](https://github.com/netascode/ansible-dc-vxlan/pull/775/files#diff-6c5e80a440f6fd5ff2080dff821a74dd3f08dde7488878a5620225a58678fdf6R79).
- The active path is `prepare_service_model` → `build_resource_data` → `manage_resources` → `NdfcModuleExecutor.execute()`. There is no later mandatory unresolved-token check.
- A current-head production-path probe captured the exact `cisco.dcnm.dcnm_fabric` arguments. With the referenced variable absent, `config[0].BGP_AUTH_KEY` remained the literal token.

A practical failure case is:

```yaml
vxlan:
  underlay:
    bgp:
      authentication_enable: true
      authentication_key_type: 3
      authentication_key: env_var_BGP_AUTH_KEY
```

If `env_var_BGP_AUTH_KEY` is absent, validation emits only a warning and the module payload remains:

```yaml
BGP_AUTH_KEY: env_var_BGP_AUTH_KEY
```

The active `dcnm_fabric` path submits that literal value to NDFC as the authentication key instead of stopping before mutation.

**Existing PR overlap**

The behavior is acknowledged by the author in [the response to the credential-plugin review](https://github.com/netascode/ansible-dc-vxlan/pull/775#issuecomment-5115409069) and in the guide, but no existing review identifies the safety consequence or requests fail-closed behavior.

**Impact**

A missing CI/CD secret, misspelled variable, or differently scoped environment can configure a predictable literal such as `env_var_BGP_AUTH_KEY`, submit invalid authentication material, or partially disrupt BGP, multicast, AAA, and other control-plane services. A warning can easily be lost in automation output while the play continues.

**Suggested fix**

Fail validation before rendering or controller mutation when any referenced variable is absent, aggregate every missing token with its model path, and repeat the guard immediately before module execution. Never return the original token as executable module data. Add exact and embedded-token tests for present, empty, missing, and removed-after-validation variables.

plugins/action/dtc/build_resource_data.py
```diff
@@ -329,6 +333,15 @@ def _build_resource(self, resource_name, rt, template):
                     module_data = hook_result['module_data']
                     break
 
+        resolved_data = copy.deepcopy(module_data)
```
mikewiebe, line 336: ### High: Resolved plaintext secrets escape into facts, a persistent multisite cache, and debug logs

**Issue**

The implementation resolves secrets into `resource_data.module_data`, returns that structure from the action plugin, registers it as `build_result`, and copies it to the `resource_data` fact. On the active deferred MSD/MCFG path it also serializes that resolved structure to `_msite_overlay_cache.json` without setting a restrictive mode or deleting it at the end of the run. NaC's policy comparison and the underlying collection can additionally log resolved values.

**Evidence**

- [`build_resource_data` deep-copies and resolves `module_data`, then stores the plaintext copy in `resource_entry`](https://github.com/netascode/ansible-dc-vxlan/pull/775/files#diff-304a2650706a45eb7ee78b114b5bdd7e6c76c3716b8239b9b1aa5e513c161657R336). Its `ActionModule.run()` returns the builder result unchanged.
- The active common role registers the action result and executes `set_fact: resource_data: "{{ build_result.resource_data }}"`; neither task has `no_log` or result sanitization.
- A current-head production `_build_resource()` probe rendered a placeholder-bearing fabric file and proved the saved YAML `data` retained the token while the returned `resource_data.fabric.module_data` contained the resolved plaintext fixture secret.
- Deferred MSD/MCFG overlay handling caches all `new_resource_data`, including resolved `module_data`, at [`pipeline_base.py` lines 697–706](https://github.com/netascode/ansible-dc-vxlan/blob/1e894e8e47749f50663cdf984c061212196a5bc7/plugins/plugin_utils/pipeline_base.py#L697-L706). A current-head production `_msite_build_overlay()` probe wrote the resolved sentinel into `_msite_overlay_cache.json`, left it present after the method returned, and created it as mode `0644` under the review host's `022` umask. Cleanup occurs only at the start of a later common-phase overlay build at [`build_resource_data.py` lines 485–489](https://github.com/netascode/ansible-dc-vxlan/blob/1e894e8e47749f50663cdf984c061212196a5bc7/plugins/action/dtc/build_resource_data.py#L485-L489), so the plaintext cache remains after the current run.
- This is a secret-bearing active path: both MSD/MCFG VRF templates render child `bgp_password` at line 41, and their deferred network/VRF-attachment templates accept freeform strings.
- NaC logs the first 80 characters of resolved desired policy values and controller `nvPairs` at `-vvv` in [`manage_resources.py` lines 610–617](https://github.com/netascode/ansible-dc-vxlan/blob/1e894e8e47749f50663cdf984c061212196a5bc7/plugins/action/dtc/manage_resources.py#L610-L617).
- At both the declared `cisco.dcnm` minimum tag `3.12.1` (`6311b66ff02b8ed593cbbfed4abdd502607dd6d2`) and current `develop` (`939c75bb667498631e9047ff9f62f424f00b98c2`), `dcnm_fabric.config` and `dcnm_policy.config` are generic lists of dictionaries with no nested `no_log` contract.
- For `dcnm_fabric`, `RestSend.__init__` logs full parameters/config, `Sender.commit()` logs the JSON payload, and `RestSend` logs the result/response at both exact DCNM refs. `dcnm_policy` does not use this same `RestSend`/`Sender` path; its proven exposure is its unmasked generic config/invocation/results plus NaC's own `-vvv` comparison.
- The author explicitly confirms that generic `env_var_` secrets remain visible in Ansible logs in [issue comment 5115364837](https://github.com/netascode/ansible-dc-vxlan/pull/775#issuecomment-5115364837). The PR guide does not disclose this limitation.

A practical MSD example uses this harmless sentinel in place of a real encrypted password:

```bash
export env_var_BLUE_BGP_PASSWORD='PR775_FAKE_SECRET'
```

```yaml
vxlan:
  fabric:
    name: MSD-FABRIC
    type: MSD
  multisite:
    overlay:
      vrfs:
        - name: PROD
          child_fabrics:
            - name: DC1
              bgp_password: env_var_BLUE_BGP_PASSWORD
              bgp_password_encryption_type: 3
```

The primary rendered YAML still appears protected:

```yaml
bgp_password: env_var_BLUE_BGP_PASSWORD
```

However, `build_resource_data` returns this resolved structure, and the common role copies it into a fact:

```yaml
resource_data:
  vrfs:
    module_data:
      - vrf_name: PROD
        child_fabric_config:
          - fabric: DC1
            bgp_password: PR775_FAKE_SECRET
```

The deferred MSD path then writes the same resolved value to `roles/dtc/common/files/msd/MSD-FABRIC/_msite_overlay_cache.json`:

```json
{
  "resource_data": {
    "vrfs": {
      "module_data": [{
        "child_fabric_config": [{
          "bgp_password": "PR775_FAKE_SECRET"
        }]
      }]
    }
  }
}
```

The tested cache was mode `0644` and remained after the method completed. Thus, although the operator sees only a token in the primary rendered file, Ansible/AAP result data and facts can retain the resolved credential, while workspace backups, artifact collection, or another local user on a shared automation host can read it from the cache.

**Existing PR overlap**

The author raised the exposure in the comment above, but it remains implemented and was not converted into an unresolved change request. The two current approvals do not address it.

**Impact**

Credentials can be retained in callback/event output, registered-variable serialization, host facts, failure results, automation-controller artifacts, a normally `0644` cache, NaC `-vvv` output, and configured `cisco.dcnm` log files. Keeping the primary rendered YAML tokenized does not provide meaningful secret protection if the same plaintext is persisted and logged elsewhere.

**Suggested fix**

Keep the model, rendered files, structural diff, `build_result`, facts, and multisite cache tokenized. Move resolution to the final module boundary, after resource selection, using a short-lived copy:

```python
execution_config = copy.deepcopy(tokenized_config)
secret_values = resolve_env_vars_strict(execution_config)
try:
    result = execute_dcnm_module(config=execution_config)
    return redact_result(result, secret_values)
finally:
    execution_config = None
```

In practice:

- Fail before invoking DCNM if any token is unresolved.
- Never return the execution copy through `invocation`, success/failure results, exceptions, or `PipelineRunnerBase.step_results`; return only recursively redacted data and use `no_log` as defense in depth where a task can still carry the copy.
- Cache only the tokenized representation, create the cache as `0600`, and delete it in a `finally` cleanup path after create/remove on both success and failure.
- Remove NaC's value-bearing policy debug output. Because NaC `no_log` does not protect `cisco.dcnm`'s own Python log files, also stop logging complete parameters, payloads, results, and responses there or centrally redact them before logging.
- Add sentinel tests proving that the resolved value reaches the mocked DCNM call but never appears in returned results, facts, cache files, display output, exceptions, or base-collection logs.

With this design, persistent state continues to contain `env_var_BLUE_BGP_PASSWORD`; only the temporary in-memory controller request contains `PR775_FAKE_SECRET`, and the returned result is redacted before the pipeline stores it.

docs/Handling_Secrets_Guide.md
```diff
@@ -0,0 +1,142 @@
...
+## Environment Variable Lookup (`env_var_`)
+
+The `env_var_` prefix instructs NaC to resolve a value from an environment variable at runtime. This works for:
+
+- **Standalone values** - the entire YAML value is an reference to an environment variable
+- **Embedded secrets** - `env_var_` references inside larger strings (e.g. freeform configuration blocks)
+
+### How It Works
+
+Any string value in the data model that contains `env_var_` is resolved by NaC VXLAN. The secret is replaced with the value of the corresponding environment variable.
```
mikewiebe: ### Medium: The guide labels plaintext substitutions as encrypted keys, and the source contract does not define `env_var_`

**Issue**

The resolver performs string substitution; it does not encrypt the substituted value. However, the guide tells users to export plaintext secrets and pair them with encryption type `3` or CLI `key 7`, both of which declare that the supplied value is already encrypted ciphertext. The PR also makes `env_var_\w+` reserved syntax in every model string, while the authoritative `nac-vxlan` source model does not define that syntax or its behavior.

**Evidence**

- The resolver applies to any string containing the prefix, not only known secret fields, as documented at [`Handling_Secrets_Guide.md` line 74](https://github.com/netascode/ansible-dc-vxlan/pull/775/files#diff-6c5e80a440f6fd5ff2080dff821a74dd3f08dde7488878a5620225a58678fdf6R74).
- At exact model ref `f997000b00b383725a9853c7f200b3a2c38cbb08`, no schema, default, documentation, or example contains `env_var_`, although many fields accepted by that schema are used as secrets.
- The guide exports plaintext `MyBgpP@$$w0rd` and `DciP@ssword`, then uses them with encryption type `3` at lines 84-112. It likewise documents a plaintext multicast key and renders `tacacs-server key 7 TacacsSecret!` at lines 122-133, even though `7` denotes an already encoded Cisco type-7 value.
- The authoritative model marks DCI eBGP password (`schemas/schema.yaml:49`), underlay BGP authentication key (`:761-762`), and multisite child BGP password (`:891-892`) as encrypted and publishes ciphertext examples (`docs/templates/vxlan/underlay/underlay_bgp.md:7-27` and `docs/templates/vxlan/overlay/overlay_vrf.md:39-40`).
- The NaC template maps the model values directly to `BGP_AUTH_KEY_TYPE` and `BGP_AUTH_KEY` (`dc_vxlan_fabric_protocols.j2:42-43`); there is no encryption step between resolution and the DCNM module payload.
- At both exact DCNM refs, `dcnm_fabric` describes `BGP_AUTH_KEY` as encrypted based on type (line 237), `PIM_HELLO_AUTH_KEY` as 3DES encrypted (line 934), `MS_IFC_BGP_PASSWORD` as an encrypted hex string (line 1461), and `BFD_AUTH_KEY` as an encrypted SHA1 secret (line 182). The resolver substitutes strings; it does not encrypt plaintext.
- The example repository at `73708666b6ce73fefbd8cb3a614efd20dddf440c` correctly represents the current released pair (`cisco.nac_dc_vxlan 0.8.1`, `cisco.dcnm 3.12.1`). Because this PR is labeled for `0.9.0`, its future companion example is a release follow-up, not a defect in the current example ref.

Practical BGP example from the guide:

```bash
export env_var_BGP_AUTH_KEY='MyBgpP@$$w0rd'
```

```yaml
vxlan:
  underlay:
    bgp:
      authentication_enable: true
      authentication_key_type: 3
      authentication_key: env_var_BGP_AUTH_KEY
```

The resolver/template path produces the equivalent module data:

```yaml
BGP_AUTH_KEY_TYPE: 3
BGP_AUTH_KEY: MyBgpP@$$w0rd
```

Type `3` tells the receiving contract that `BGP_AUTH_KEY` is already 3DES-encrypted, but the value is still the original plaintext because the resolver only replaced the token. In practical terms, this is like setting `value_is_encrypted: true` while passing an unencrypted password: NDFC can reject or misinterpret it, or configure authentication material different from what the operator intended. The guide's `tacacs-server key 7 TacacsSecret!` example has the same mismatch—`7` says the following value is already type-7 encoded, while `TacacsSecret!` is plaintext.

There is also a separate source-contract problem. For example, a user could legitimately place this in a freeform model field:

```yaml
banner_freeform: |
  Contact env_var_SUPPORT_TEAM after hours.
```

Because resolution is global, that ordinary sentence is rewritten if `env_var_SUPPORT_TEAM` exists and emits a missing-variable warning if it does not. Broad embedded substitution may be intentional, but then the source model must define the prefix as reserved syntax, its supported locations, missing-variable behavior, rotation behavior, value-format rules, and a way to escape literal `env_var_...` text.

**Existing PR overlap**

[Review 4915997413](https://github.com/netascode/ansible-dc-vxlan/pull/775#pullrequestreview-4915997413) explicitly requests documentation “in netascode.” The subsequent commit added only this repository's guide, so the source-repository portion remains unanswered.

**Impact**

An operator can follow the guide exactly and submit plaintext where NDFC expects ciphertext matching the selected encryption type, causing rejection, misinterpretation, or unintended authentication configuration. Independently, ordinary model text can be rewritten merely because it contains the reserved-looking prefix. The authoritative source model and the solution behavior therefore disagree about both accepted value format and syntax/lifecycle semantics.

**Suggested fix**

Update `nac-vxlan` guide change before `0.9.0` that defines the prefix as reserved syntax, the supported fields/locations, embedded-token behavior, missing-variable behavior, rotation semantics, required value formats, and how users escape literal `env_var_...` text.

For encrypted fields, document one explicit contract:

**Ciphertext-in:** require the environment variable itself to contain ciphertext encoded for the selected type, for example `export env_var_BGP_AUTH_KEY='<3DES-ciphertext>'` when `authentication_key_type: 3` is selected.

Correct the BGP, DCI, multicast, BFD, and AAA/type-7 examples accordingly. If substitution is intended only for known secret fields, constrain the implementation to those model paths; if broad substitution is intentional, document and test an escape mechanism. Add the `0.9.0` example-repository companion and tests that assert the exact final payload and encoding expected by DCNM.

plugins/action/common/prepare_plugins/prep_005_resolve_env_vars.py
marehler: Addressed, the resolver now fails on missing environment variables.
Previously an unset env_var_* token was warning-only and the literal token was carried forward to NDFC as configuration. That's fixed at two levels:

1. Fail-closed at validation (prep_005_resolve_env_vars.py)
- validate_env_vars_recursive() now aggregates every missing token together with its data-model path instead of emitting per-token warnings.
- PreparePlugin.prepare() sets results['failed'] = True with a message listing all missing variables and their locations. Because this runs in prepare_service_model during the validate stage, the run stops before any rendering or controller mutation. This uses the framework's existing fail-closed convention (same pattern as prep_999_verify).

2. Defense-in-depth at resolution time
- resolve_env_var_token() now raises UnresolvedEnvVarError instead of returning the literal token. This covers the "removed-after-validation" race: if a variable is unset/removed between validate and the runtime resolvers (build_resource_data.py, manage_resources.py), the play fails loudly rather than silently emitting env_var_* as the credential value. A literal token can no longer become executable module data on any path.

Docs: Handling_Secrets_Guide.md updated, the previous "a warning is displayed and the secret is left unchanged" line now documents the failure behavior.
This directly resolves the flagged case: with env_var_BGP_AUTH_KEY absent, the run now stops during validate instead of submitting BGP_AUTH_KEY: env_var_BGP_AUTH_KEY to NDFC.

docs/Handling_Secrets_Guide.md
marehler: Fixed the guide. Added an explicit description that "env_var_ substitutes only, it does not encrypt" contract. Corrected the BGP/DCI/multicast/BFD and AAA type-7 examples to a ciphertext-in model (the environment variable must hold the value already encoded for the selected encryption type). Also documented the global env_var_\w+ substitution scope, the rewrite of ordinary text containing the prefix, and how to keep literal env_var_... text. Broad substitution is intentional and left as is.

plugins/action/dtc/build_resource_data.py
marehler: Partially addressed. Applied in-repo hardening:
* MSD overlay cache (pipeline_base.py): now written 0600 and deleted immediately after the consuming REMOVE pipeline loads it (and on load failure), so plaintext no longer persists after the run.
* -vvv policy diff log (manage_resources.py): no longer emits values — logs only the differing key name and set/empty/missing presence.
* Guide: Documented logging exposure in limitations section

Considered to add "no_log: true" in the build_resource_data / manage_resources / remove_resources role tasks. However, that hides the entire task results, complicating debugging, and doesn't cover cisco.dcnm's logging. Please let me know if you want me to implement the "no_log".

Moving the resolution to module boundary + redact returned results/facts/cache would be a core pipeline redesign. Addressing the cisco.dcnm payload logging would require to change the cisco.dcnm collection.

marehler:
> @marehler Given that we are recursively walking the entire model_extended data model I am concerned about the performance impact at scale. Have you scaled the data model (above 10MB) with a large number of env_vars* to profile the performance impact?

I've tested with a data model including 30k VRFs, 30k networks, 1000 env_vars_ that is ~ 10MB. The impact seems to be negligible. 

Results with develop branch:
````
TASKS RECAP **************************************************************************************************************************************************
Dienstag 18 August 2026  15:56:28 +0200 (0:00:00.163)       0:01:51.518 ******* 
=============================================================================== 
cisco.nac_dc_vxlan.validate : Perform Required Syntax and Semantic Model Validation ------------------------------------------------------------------ 44.00s
cisco.nac_dc_vxlan.validate : Perform Custom Enhanced Syntax and Semantic Model Validation ----------------------------------------------------------- 41.15s
cisco.nac_dc_vxlan.validate : Prepare Service Model -------------------------------------------------------------------------------------------------- 10.28s
cisco.nac_dc_vxlan.validate : Store Golden Service Model Data ----------------------------------------------------------------------------------------- 4.74s
cisco.nac_dc_vxlan.validate : Store Extended Service Model Data --------------------------------------------------------------------------------------- 4.07s
cisco.nac_dc_vxlan.validate : Merge factory and custom defaults --------------------------------------------------------------------------------------- 3.56s
````
Results with env_var_ mechanism:
````
TASKS RECAP **************************************************************************************************************************************************
Dienstag 18 August 2026  16:00:28 +0200 (0:00:00.160)       0:01:52.630 ******* 
=============================================================================== 
cisco.nac_dc_vxlan.validate : Perform Required Syntax and Semantic Model Validation ------------------------------------------------------------------ 44.02s
cisco.nac_dc_vxlan.validate : Perform Custom Enhanced Syntax and Semantic Model Validation ----------------------------------------------------------- 41.99s
cisco.nac_dc_vxlan.validate : Prepare Service Model -------------------------------------------------------------------------------------------------- 10.55s
cisco.nac_dc_vxlan.validate : Store Golden Service Model Data ----------------------------------------------------------------------------------------- 4.76s
cisco.nac_dc_vxlan.validate : Store Extended Service Model Data --------------------------------------------------------------------------------------- 4.06s
cisco.nac_dc_vxlan.validate : Merge factory and custom defaults --------------------------------------------------------------------------------------- 3.53s

````

marehler:
@mikewiebe, I've addressed most of your comments. The main open issue is that resolved values are still visible in verbose logs. For logs generated by the NaC collection, we could add "no_log: true" in the build_resource_data, manage_resources, and remove_resources role tasks. For logs generated by the cisco.dcnm collection modules, suppressing the resolved values would require changes in those modules. For now, I've documented this as a limitation.

marehler:
Tested with ND 4.3.1. Works fine.

mikewiebe (approved):

PR #784, opened 2026-04-29
Refactor - Diff policies

ccoueffe:
## Related Collection Role

* [ ] other

## Related Data Model Element

* [ ] vxlan.fabric
* [ ] vxlan.global
* [ ] vxlan.topology
* [ ] vxlan.underlay
* [ ] vxlan.overlay
* [x] vxlan.policy
* [ ] vxlan.multisite
* [ ] defaults.vxlan
* [ ] other

## Test Notes

force_run_all: True
With change:

```
───────────────────────────────────────────────────────────────────────────────
CREATE [nac-fabric1] Step 20/21: policy (_policy_remote_diff)
───────────────────────────────────────────────────────────────────────────────
POLICY DIFF [nac_alias]: var CONF desired='cli alias name wr copy running startup\ncli alias name sr show running' vs ctrl='cli alias name wr copy running startup\ncli alias name sr show running-config'
POLICY DIFF [nac_banner]: var BANNER desired='***************************************************************************\n*   ' vs ctrl='***************************************************************************\n*   '
POLICY DIFF [nac_grpc_config]: var CONF desired='grpc gnmi max-concurrent-calls 16\ngrpc use-vrf default' vs ctrl='grpc gnmi max-concurrent-calls 15\ngrpc use-vrf default'
CREATE [nac-fabric1] policy controller diff: 8 desired, 9 on controller -> 3 to push
CREATE [nac-fabric1] policy → ok (changed=False) [1.2s]

───────────────────────────────────────────────────────────────────────────────
CREATE [nac-fabric1] Step 21/21: policy (dcnm_policy)
───────────────────────────────────────────────────────────────────────────────
CREATE [nac-fabric1] Executing dcnm_policy for policy with state=merged, save=None, deploy=False, items=1
CREATE [nac-fabric1] policy → ok (changed=True) [3.2s]
```

Without :

```
───────────────────────────────────────────────────────────────────────────────
CREATE [nac-fabric1] Step 20/21: policy (_policy_remote_diff)
───────────────────────────────────────────────────────────────────────────────
CREATE [nac-fabric1] policy controller diff: 8 desired, 9 on controller -> 0 to push
CREATE [nac-fabric1] policy → ok (changed=False) [1.3s]

───────────────────────────────────────────────────────────────────────────────
CREATE [nac-fabric1] Step 21/21: policy (dcnm_policy)
───────────────────────────────────────────────────────────────────────────────
CREATE [nac-fabric1] policy → skipped (no diff) [0.0s]
```

Force_run_all: False
And one change:

```
───────────────────────────────────────────────────────────────────────────────
CREATE [nac-fabric1] Step 20/21: policy (_policy_remote_diff)
───────────────────────────────────────────────────────────────────────────────
CREATE [nac-fabric1] policy → ok (changed=False) [0.0s]

───────────────────────────────────────────────────────────────────────────────
CREATE [nac-fabric1] Step 21/21: policy (dcnm_policy)
───────────────────────────────────────────────────────────────────────────────
CREATE [nac-fabric1] Executing dcnm_policy for policy with state=merged, save=None, deploy=False, items=1
CREATE [nac-fabric1] policy → ok (changed=True) [3.2s]
```

In ndfc_policy_comparison.yml:

```
comparison_summary:
  timestamp: '2026-05-06T11:51:09.342029'
  source_file: !!python/object/apply:ansible.utils.unsafe_proxy.AnsibleUnsafeText
  - /home/cisco/collections/ansible_collections/cisco/nac_dc_vxlan/roles/dtc/common/files/vxlan/nac-fabric1/ndfc_policy.yml
  total_updated: 3
  total_removed: 0
  total_equal: 5
updated_items:
- switch:
  - ip: 10.229.42.155
    policies:
    - create_additional_policy: false
[..]
```

## Checklist

### DIFF
--- a/plugins/action/dtc/diff_compare.py
+++ b/plugins/action/dtc/diff_compare.py
@@ -84,10 +84,14 @@ class ActionModule(ActionBase):
         # Normalize omit placeholder strings between old and new items
         old_items, new_items = self.normalize_omit_placeholders(old_items, new_items)
 
-        updated_items, removed_items, equal_items = self.compare_items(old_items, new_items)
-
-        if self.new_file_path.endswith('ndfc_interface_all.yml'):
+        # File-type specific comparison
+        if self.new_file_path.endswith('ndfc_policy.yml'):
+            updated_items, removed_items, equal_items = self.compare_policies(old_items, new_items)
+        elif self.new_file_path.endswith('ndfc_interface_all.yml'):
+            updated_items, removed_items, equal_items = self.compare_items(old_items, new_items)
             removed_items = self.order_interface_remove(removed_items)
+        else:
+            updated_items, removed_items, equal_items = self.compare_items(old_items, new_items)
 
         results['compare'] = {"updated": updated_items, "removed": removed_items, "equal": equal_items}
 
@@ -96,6 +100,15 @@ class ActionModule(ActionBase):
 
         return results['compare']
 
+    @staticmethod
+    def _count_policies(switch_blocks):
+        """Count individual policies across nested switch blocks."""
+        count = 0
+        for sb in switch_blocks:
+            for sw in sb.get('switch', []):
+                count += len(sw.get('policies', []))
+        return count
+    
     def write_comparison_results(self, compare_results):
         """
         Write comparison results to a unique file in the same directory as new_file_path.
@@ -120,9 +133,16 @@ class ActionModule(ActionBase):
             'comparison_summary': {
                 'timestamp': datetime.datetime.now().isoformat(),
                 'source_file': self.new_file_path,
-                'total_updated': len(compare_results.get('updated', [])),
-                'total_removed': len(compare_results.get('removed', [])),
-                'total_equal': len(compare_results.get('equal', []))
+                'total_updated': self._count_policies(compare_results.get('updated', []))
+                    if self.new_file_path.endswith('ndfc_policy.yml')
+                    else len(compare_results.get('updated', [])),
+                'total_removed': self._count_policies(compare_results.get('removed', []))
+                    if self.new_file_path.endswith('ndfc_policy.yml')
+                    else len(compare_results.get('removed', [])),
+                'total_equal': self._count_policies(compare_results.get('equal', []))
+                    if self.new_file_path.endswith('ndfc_policy.yml')
+                    else len(compare_results.get('equal', [])),
+            },
             },
             'updated_items': compare_results.get('updated', []),
             'removed_items': compare_results.get('removed', []),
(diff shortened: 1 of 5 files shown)


plugins/action/dtc/diff_compare.py
```diff
@@ -120,9 +133,16 @@ def write_comparison_results(self, compare_results):
...
+                'total_updated': self._count_policies(compare_results.get('updated', []))
+                    if self.new_file_path.endswith('ndfc_policy.yml')
+                    else len(compare_results.get('updated', [])),
+                'total_removed': self._count_policies(compare_results.get('removed', []))
+                    if self.new_file_path.endswith('ndfc_policy.yml')
+                    else len(compare_results.get('removed', [])),
+                'total_equal': self._count_policies(compare_results.get('equal', []))
+                    if self.new_file_path.endswith('ndfc_policy.yml')
+                    else len(compare_results.get('equal', [])),
+            },
```
mikewiebe, line 152: @ccoueffe something looks off here

The PR does not compile. py_compile fails with IndentationError: unexpected indent. output_data is closed early around line 146, leaving updated_items, removed_items, and equal_items outside the dict. compare_policies is also misindented at line 282. This blocks the action plugin from loading.

Is there an extra `},`?

Image

plugins/action/dtc/manage_resources.py
```diff
@@ -206,6 +208,210 @@ def _fabric_links_query_and_filter(self, resource_name, step):
...
+            FABRIC_NAME, POLICY_ID, POLICY_DESC
+
+        Args:
+            resource_name: Resource identifier (policy).
+            desired_config: Current rendered policy config.
+
+        Returns:
+            dict with changed and optionally failed/msg keys.
+        """
+        api_path = (
```
mikewiebe: The controller diff calls policies/pagination?fabricName=... without page size, offset, or a pagination loop. If that endpoint returns a default page, the diff sees partial controller state and will push many unchanged policies. If it returns all, this is still broader than the existing dcnm_policy path, which queries policies only for target switches.

Suggested fix instead of using a query for all policies in the fabric

**Approach 1: Preferred**

Use the same bounded query shape that cisco.dcnm.dcnm_policy already uses:

`/appcenter/cisco/ndfc/api/v1/lan-fabric/rest/control/policies/switches?serialNumber=<comma-separated target serials>
`

Why: the desired policy config is already scoped to specific switches, so querying all policies for the whole fabric is unnecessary. Build the serial list from the data model switch inventory, or from controller inventory if needed, then compare only policies on those switches. This mirrors the existing module behavior and **scales better**.

This approach should scale better in the common case because it bounds the controller diff to the switches NAC is actually managing.

The cost changes from roughly:

`all policies in fabric`

to:

`policies on target switches only`

So if a fabric has 500 switches but the play is touching 20, querying /policies/switches?serialNumber=... is much friendlier than fabric-wide policies/pagination?fabricName=....

The caveat: if each target switch has a very large number of policies, approach 1 still has to retrieve and compare all policies for those switches. It reduces unrelated data, but it does not solve high policy density per switch.

I would address it like this:

Use /policies/switches?serialNumber=<comma-separated target serials> to match dcnm_policy behavior.
Chunk serials if the serial list can make the URL too long.
Build desired policy keys first, then index remote policies in one pass.
Avoid nested comparisons like “for each desired policy, scan every remote policy”.


**Only fall back to fabric-wide pagination (Approach 2) if no switch serials are available.**

**Approach 2**

If the pagenation approach is really needed here then add a pagination helper

```python
from urllib.parse import urlencode

def _fetch_all_policy_pages(self, page_size=1000):
    policies = []
    offset = 0

    while True:
        query = urlencode({
            "fabricName": self.fabric_name,
            "max": page_size,
            "offset": offset,
        })
        path = (
            "/appcenter/cisco/ndfc/api/v1/lan-fabric/rest/control/"
            f"policies/pagination?{query}"
        )

        result = self.executor.execute_rest("GET", path)
        if result.get("failed"):
            return result

        response = result.get("response", {})
        data = response.get("DATA", []) if isinstance(response, dict) else response

        if not data:
            break

        policies.extend(data)

        if len(data) < page_size:
            break

        offset += page_size

    return {"failed": False, "policies": policies}
```

plugins/action/dtc/manage_resources.py
```diff
@@ -206,6 +208,210 @@ def _fabric_links_query_and_filter(self, resource_name, step):
 
         return {'changed': False}
 
+   def _policy_remote_diff(self, resource_name, step):
```
mikewiebe: This is `def` is indented incorrectly 

Image

plugins/action/dtc/diff_compare.py
ccoueffe: removed

plugins/action/dtc/manage_resources.py
ccoueffe: fix

plugins/action/dtc/manage_resources.py
ccoueffe: remove pagination

mtarking (approved):

PR #789, opened 2026-05-03
Improve role validate performance

mikewiebe:
## Related Collection Role

* [ ] other

## Related Data Model Element

* [x] vxlan.fabric
* [x] vxlan.global
* [x] vxlan.topology
* [x] vxlan.underlay
* [x] vxlan.overlay
* [x] vxlan.policy
* [x] vxlan.multisite
* [ ] defaults.vxlan
* [ ] other

### Summary

This PR adds several validate/run-map performance optimizations for NAC VXLAN runs, for both targeted runs with `force_run_all: false` and full runs `force_run_all: true`.

The main improvement is avoiding expensive model loading, model diff generation, and full data-model argument passing when they are not needed. Functional behavior is preserved, while no-change targeted runs can short-circuit faster.

### Performance Updates

- Reused the loaded validation data model across semantic rule checks instead of reloading it repeatedly.
- Ensured syntax validation runs only once per validation execution when a schema is present.
- Added lightweight `fabric_name` support to run-map action plugins so tasks do not need to pass the full extended data model just to update/read run-map state.
- Added validate runtime mode flags:
  - `validate_checksum_compare_enabled`
  - `validate_model_diff_enabled`
- Replaced expensive no-change model comparisons with checksum-based comparisons when model diff display is disabled.
- Skipped `include_vars` loading of previous/current model JSON files unless `display_model_diff: true` and the run mode actually requires diff display.
- Preserved optional model diff display behavior behind `display_model_diff`.
- Narrowed cleanup so it only removes previous model JSON files instead of matching all service model JSON files.
- Kept `force_run_all: true` behavior as a full run path and optimized `force_run_all: false` for targeted/no-change detection.

### Details Hot Areas

Screenshot 2026-05-03 at 7 39 51 PM

### Behavior Notes

- `display_model_diff: false` avoids the expensive model diff display path.
- `display_model_diff: true` still allows model changes to be displayed when running in a diff-enabled targeted mode.
- `force_run_all: true` bypasses the targeted checksum short-circuit behavior, as expected.
- The run-map plugins remain backward compatible by falling back to `data_model.vxlan.fabric.name` if `fabric_name` is not provided.

### Validation

Validated with the full re-resync playbook:

```bash
source env.sh ; ansible-playbook -i inventory.yaml vxlan.yaml

Playbook result: success
Total runtime: 1:09:07
Recap: ok=48 changed=4 unreachable=0 failed=0 skipped=18
```

### Testbed Scale

Screenshot 2026-05-03 at 6 22 29 PM

## Checklist

### DIFF
--- a/roles/validate/tasks/manage_model_files_current.yml
+++ b/roles/validate/tasks/manage_model_files_current.yml
@@ -39,11 +39,26 @@
     force: true
   delegate_to: localhost
 
+- name: Stat Current Golden Service Model Data
+  ansible.builtin.stat:
+    path: "{{ role_path }}/files/{{ data_model_extended.vxlan.fabric.name }}_service_model_golden.json"
+  register: current_golden_stat
+  when: validate_checksum_compare_enabled | default(false) | bool
+  delegate_to: localhost
+
+- name: Stat Current Extended Service Model Data
+  ansible.builtin.stat:
+    path: "{{ role_path }}/files/{{ data_model_extended.vxlan.fabric.name }}_service_model_extended.json"
+  register: current_extended_stat
+  when: validate_checksum_compare_enabled | default(false) | bool
+  delegate_to: localhost
+
 # Read current golden service model data into a variable called 'smd_golden_current'
 - name: Read Current Service Model Data from Host
   ansible.builtin.include_vars:
     file: "{{ role_path }}/files/{{ data_model_extended.vxlan.fabric.name }}_service_model_golden.json"
   register: smd_golden_current
+  when: validate_model_diff_enabled | default(false) | bool
   delegate_to: localhost
 
 # Read current extended service model data into a variable called 'data_model_extended_current'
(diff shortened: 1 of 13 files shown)


roles/validate/tasks/manage_model_files_current.yml
```diff
@@ -39,67 +39,87 @@
...
-    before: "{{ data_model_extended_previous }}"
-    after: "{{ data_model_extended_current }}"
+    before: "{{ data_model_extended_previous | default({}) }}"
+    after: "{{ data_model_extended_current | default({}) }}"
   register: data_model_extended_diff
-  when: check_roles['save_previous']
+  when:
+    - validate_model_diff_enabled | default(false) | bool
+    - extended_stat.stat.exists | default(false)
   delegate_to: localhost
```
mtarking, line 90: Can we put `verbosity: 2` or `verbosity: 3` on these tasks?

mtarking (changes requested):

roles/validate/tasks/manage_model_files_current.yml
mikewiebe: Will do this in an incremental update

PR #794, opened 2026-05-06
Refactor - enhance diff controller links

ccoueffe:
## Related Collection Role

* [ ] other

## Related Data Model Element

* [ ] vxlan.fabric
* [ ] vxlan.global
* [ ] vxlan.topology
* [ ] vxlan.underlay
* [ ] vxlan.overlay
* [ ] vxlan.policy
* [ ] vxlan.multisite
* [ ] defaults.vxlan
* [ ] other

## Proposed Changes

Fabric Links: Pre-Filtering and Diff Optimization

This enhancement adds a pre-filtering and diff step for fabric links before invoking dcnm_links, reducing unnecessary API calls in large-scale fabrics (100+ switches, 200+ links).

- 4-step pipeline:
  1. Resolve desired fabric links from the data model (respecting diff_run mode)
  2. Query all existing links from NDFC controller (dcnm_links state: query)
  3. Enrich desired links with serial numbers and resolve hostnames via existing_links_check
  4. Post-enrichment diff — compare enriched links field-by-field against controller state
- Field-level comparison: compares template name and profile fields (MTU, admin state, peer descriptions, freeform config, IPv4 addresses, MACsec) against NDFC nvPairs
- Template normalization awareness: accounts for NDFC auto-upgrading int_pre_provision_intra_fabric_link to int_intra_fabric_num_link / int_intra_fabric_unnum_link when switches come online — these transitions are not flagged as changes
- Bidirectional matching: builds a lookup keyed by (switch, interface) pairs in both directions so link A→B matches B→A on the controller regardless of how NDFC stores it
- Result: only new or modified links are sent to dcnm_links with state: merged, skipping links that already match the desired state on the controller

## Checklist

### DIFF
--- a/plugins/action/dtc/manage_resources.py
+++ b/plugins/action/dtc/manage_resources.py
@@ -190,97 +190,20 @@ class ResourceManager(PipelineRunnerBase):
                 ),
             }
 
-        # Step 4: Post-enrichment diff — compare enriched links against controller
+        # Step 4: Update resource data with filtered links
+        # Value-level profile comparison is now done inside existing_links_check
+        # during endpoint matching — no second pass needed.
         required_links = filter_result.get('required_links', [])
-
-        # Build lookup of existing links: {(sw1_ip_or_name, if1, sw2_ip_or_name, if2): nvPairs}
-        existing_lookup = {}
-        for el in existing_links:
-            sw1 = el.get('sw1-info', {})
-            sw2 = el.get('sw2-info', {})
-            key_fwd = (
-                sw1.get('sw-sys-name', '').lower(),
-                sw1.get('if-name', '').lower(),
-                sw2.get('sw-sys-name', '').lower(),
-                sw2.get('if-name', '').lower(),
-            )
-            key_rev = (key_fwd[2], key_fwd[3], key_fwd[0], key_fwd[1])
-            existing_lookup[key_fwd] = el
-            existing_lookup[key_rev] = el
-
-        # Profile field -> nvPairs key mapping
-        profile_to_nv = {
-            'admin_state': 'ADMIN_STATE',
-            'mtu': 'MTU',
-            'peer1_description': 'PEER1_DESC',
-            'peer2_description': 'PEER2_DESC',
-            'peer1_cmds': 'PEER1_CONF',
-            'peer2_cmds': 'PEER2_CONF',
-            'peer1_freeform': 'PEER1_CONF',
-            'peer2_freeform': 'PEER2_CONF',
-            'peer1_ipv4_addr': 'PEER1_IP',
-            'peer2_ipv4_addr': 'PEER2_IP',
-            'enable_macsec': 'ENABLE_MACSEC',
-        }
-
-        diff_links = []
-        for link in required_links:
-            src = link.get('src_device', '').lower()
-            src_if = link.get('src_interface', '').lower()
-            dst = link.get('dst_device', '').lower()
-            dst_if = link.get('dst_interface', '').lower()
-
-            el = existing_lookup.get((src, src_if, dst, dst_if))
-            if el is None:
-                # New link — not on controller
-                diff_links.append(link)
-                continue
-
-            # Template change — skip if NDFC upgraded pre-provision to num/unnum
-            desired_tpl = link.get('template', '')
-            ctrl_tpl = el.get('templateName', '')
-            if desired_tpl and desired_tpl != ctrl_tpl:
-                # Pre-provision -> numbered/unnumbered is expected NDFC behavior
-                ndfc_upgraded = (
-                    desired_tpl == 'int_pre_provision_intra_fabric_link'
-                    and ctrl_tpl in (
-                        'int_intra_fabric_num_link',
-                        'int_intra_fabric_unnum_link',
-                    )
-                )
-                if not ndfc_upgraded:
-                    diff_links.append(link)
-                    continue
-
-            # Compare profile fields against nvPairs
-            nv = el.get('nvPairs', {})
-            profile = link.get('profile', {})
-            changed = False
-            for pkey, nv_key in profile_to_nv.items():
-                if pkey in profile:
-                    desired_val = str(profile[pkey]).strip()
-                    ctrl_val = str(nv.get(nv_key, '')).strip()
-                    # Case-insensitive for booleans
-                    if desired_val.lower() != ctrl_val.lower():
-                        display.vvv(
-                            f"LINK DIFF [{src}:{src_if}->{dst}:{dst_if}]: "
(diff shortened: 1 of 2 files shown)


plugins/action/dtc/manage_resources.py
```diff
@@ -190,18 +190,97 @@ def _fabric_links_query_and_filter(self, resource_name, step):
...
+                )
+                if not ndfc_upgraded:
+                    diff_links.append(link)
+                    continue
+
+            # Compare profile fields against nvPairs
+            nv = el.get('nvPairs', {})
+            profile = link.get('profile', {})
+            changed = False
+            for pkey, nv_key in profile_to_nv.items():
```
mikewiebe: AI flagged a potential issue here as high...

plugins/action/dtc/manage_resources.py:259-264 

only compares profile keys that are present in the raw rendered link. If a user removes `peer1_freeform/peer2_freeform`, the rendered profile no longer contains `peer1_cmds/peer2_cmds`, so this prefilter treats the link as unchanged and skips `dcnm_links`. 

Previously `dcnm_links state=replaced` would default those fields to empty and clear stale `PEER*_CONF` on the controller. The prefilter should compare managed defaults too, not only present keys.

mikewiebe: Also, this one worries me.. need to check

plugins/action/dtc/manage_resources.py:259-264 

compares `str(profile[pkey]).strip()` directly to controller nvPairs. That can differ from what dcnm_links actually sends, especially for command lists/blocks, booleans, IP canonicalization, and defaults. This risks both false positives and false negatives. The safer fix is to normalize desired values using the same shape as dcnm_links payloads before comparing.

plugins/action/dtc/manage_resources.py
```diff
@@ -190,18 +190,97 @@ def _fabric_links_query_and_filter(self, resource_name, step):
...
+                sw1.get('sw-sys-name', '').lower(),
+                sw1.get('if-name', '').lower(),
+                sw2.get('sw-sys-name', '').lower(),
+                sw2.get('if-name', '').lower(),
+            )
+            key_rev = (key_fwd[2], key_fwd[3], key_fwd[0], key_fwd[1])
+            existing_lookup[key_fwd] = el
+            existing_lookup[key_rev] = el
+
+        # Profile field -> nvPairs key mapping
```
mikewiebe: plugins/action/dtc/manage_resources.py:211-224 

only maps a subset of fields. dcnm_links also handles fields such as BFD echo disable, interface VRF, IPv6 peer IPs, and other template-specific values. Any desired change outside this map can be filtered out before the module sees it.

Will this be an issue?

mikewiebe:
Need to check this


This PR adds an O(n) lookup after cisco.nac_dc_vxlan.dtc.existing_links_check
, but existing_links_check.py still does the expensive configured-link × existing-link × switch scan first. For large fabrics, the real scale win would be moving the lookup/indexing into that plugin and doing enrichment/diff in one keyed pass.

plugins/action/dtc/manage_resources.py
ccoueffe: Fixed. Added a reverse check in `_link_profile_changed()` inside `existing_links_check.py`.

After the forward comparison (desired profile keys vs controller), we now check if the controller has non-empty `PEER1_CONF`/`PEER2_CONF` for keys absent from the desired profile. This catches the case where a user removes `peer1_freeform`/`peer2_freeform` from the data model — the rendered link no longer contains `peer1_cmds`/`peer2_cmds`, but the controller still has values that need to be cleared.

Note on the data flow: the J2 template (`ndfc_fabric_links.j2`) converts `peer1_freeform` → `peer1_cmds` during rendering. So in the `required_links` output, the profile key is always `peer1_cmds` (never `peer1_freeform`). The old `peer1_freeform`/`peer2_freeform` entries in the mapping were dead code and have been removed.

See commit 09c39e5.

ccoueffe: Investigated this. The risk is low in practice for the current NAC data flow:

- **Command lists/blocks**: The J2 template (`ndfc_fabric_links.j2`) renders `peer1_freeform` as a YAML block scalar string (`peer1_cmds`), not a list. So `str()` produces the same shape as what dcnm_links sends.
- **Booleans**: `.lower()` comparison handles `True`/`true`/`TRUE` etc.
- **IP canonicalization**: IPs in the enriched link come directly from the controller's `nvPairs` (populated by `existing_links_check` from `PEER1_IP`/`PEER2_IP`), so we're comparing the same source values.
- **Defaults**: Handled by the reverse check — if a field is absent from the desired profile but present on the controller, we flag it as changed.

A false positive (we say "changed" when it isn't) is safe — the link just gets re-evaluated by `dcnm_links` which does its own idempotent comparison. A false negative would be a bug, but the forward + reverse checks cover the cases that flow through NAC today.

plugins/action/dtc/manage_resources.py
ccoueffe: Not an issue. The mapping covers all fields that ND accepts and returns for the 3 link templates used by NAC (`int_pre_provision_intra_fabric_link`, `int_intra_fabric_num_link`, `int_intra_fabric_unnum_link`).

The J2 template (`ndfc_fabric_links.j2`) only renders: `admin_state`, `mtu`, `peer1_description`, `peer2_description`, `peer1_cmds`, `peer2_cmds`, `peer1_ipv4_addr`, `peer2_ipv4_addr`. Fields like BFD echo disable, interface VRF, or IPv6 peer IPs are not parameters of these templates — they can't be passed to or received from ND through NAC, so there's no comparison gap.

The `PROFILE_TO_NV` mapping is defined as a class constant on `existing_links_check.ActionModule`, so if the J2 template adds new fields in the future, the mapping just needs a corresponding entry.

ccoueffe:
Addressed. The value-level profile comparison has been moved into `existing_links_check.py` — it now runs during the existing endpoint matching loop, so there's no second pass.

The `_link_profile_changed()` method is called right after the link is enriched (IPs, template, MACSEC), at the point where the code already has both the desired link and the matching existing link side by side. If the profile matches the controller, the link goes to `not_required_links` instead of `required_links`.

`manage_resources.py` now just passes `required_links` directly to `module_data` — the entire prefilter block (lookup construction, second endpoint scan, template check, value comparison) has been removed.

This also benefits the YAML task path (`roles/dtc/create/tasks/common/links.yml`) which calls the same plugin.

**Performance impact**: For idempotent runs (no changes), fabric links are now skipped entirely — 0 links sent to `dcnm_links` instead of all N. For a 500-link fabric with no changes, this avoids ~500 unnecessary NDFC API calls.

See commit 09c39e5.

mikewiebe (approved):

mtarking (approved):

PR #802, opened 2026-05-28
Fix hostname policy management so the bulk policy API fallback behaves correctly

mikewiebe:
## Related Collection Role

* [x] cisco.nac_dc_vxlan.dtc.deploy
* [ ] other

## Related Data Model Element

* [ ] vxlan.fabric
* [ ] vxlan.global
* [x] vxlan.topology
* [ ] vxlan.underlay
* [ ] vxlan.overlay
* [ ] vxlan.policy
* [ ] vxlan.multisite
* [ ] defaults.vxlan
* [ ] other

## Summary

Fix hostname policy management so the bulk policy API fallback behaves correctly and duplicate hostname policies fail loudly.

## Changes

- Added explicit `NdfcBulkPolicyApiUnavailable` handling for the NDFC bulk policy lookup API.
- Updated bulk policy lookup to inspect `RETURN_CODE` from `dcnm_rest` results.
- Treat `404` from the bulk policy endpoint as “bulk API unavailable” and fall back to per-switch policy lookup.
- Treat other non-200 bulk lookup responses as real failures instead of silently continuing.
- Added `NdfcDuplicatePolicyError` for duplicate policy drift detection.
- Updated `host_11_1` lookup logic to fail when more than one hostname policy exists for the same switch serial.
- Limited fallback in `update_switch_hostname_policy.py` to only the explicit bulk-unavailable case.

## Why

On unpatched ND/NDFC systems, the bulk policy lookup endpoint can return `404`. Previously, that result was interpreted as an empty successful lookup, causing the hostname policy action to think no `host_11_1` policy existed and create a new same-priority policy. Repeated runs could create duplicate hostname policies, leaving NDFC to choose whichever policy wins.

This change makes the intended fallback path explicit and prevents duplicate hostname policies from being silently ignored.

## Verification

- Python syntax check passed for the modified files.
- Mocked `404` bulk lookup confirms fallback signal is raised.
- Mocked duplicate `host_11_1` policy lookup confirms duplicate detection fails loudly.

If we detect a duplicate policy then the following error is generated and stop execution

```bash
atal: [nac-mcfg-fabric1]: FAILED! => {
    "changed": false,
    "msg": "Manage resources failed: Duplicate host_11_1 policies found for switch serial 9G6AZ38620E. Policy IDs: POLICY-346130, POLICY-344650, POLICY-343180. Duplicate hostname policies are controller drift and must be reconciled before continuing."
}
```

## Checklist

### DIFF
--- a/plugins/action/dtc/update_switch_hostname_policy.py
+++ b/plugins/action/dtc/update_switch_hostname_policy.py
@@ -28,6 +28,7 @@ import json
 
 from ansible.plugins.action import ActionBase
 from ansible_collections.cisco.nac_dc_vxlan.plugins.plugin_utils.helper_functions import (
+    NdfcBulkPolicyApiUnavailable,
     ndfc_get_fabric_policies_by_template,
     ndfc_get_switch_policy_using_template
 )
@@ -73,8 +74,8 @@ class ActionModule(ActionBase):
                 )
                 # If successful, return the bulk result
                 return policies_dict, True
-            except Exception as e:
-                # Bulk API not available or failed, mark it and fall back
+            except NdfcBulkPolicyApiUnavailable:
+                # Bulk API not available, mark it and fall back
                 self.bulk_api_available = False
 
         # Fallback: Query each switch individually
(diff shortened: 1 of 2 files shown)


juarocha (approved):
LGTM, tested and work as expected

PR #803, opened 2026-05-29
manage edge bgp_password

ccoueffe:
update bgp_password_enable for edge connection. Change to false. By default it was enabled but if you don't provide password it failed, because password can't be empty.

## Related Collection Role

* [ ] other

## Related Data Model Element

* [ ] vxlan.fabric
* [ ] vxlan.global
* [x] vxlan.topology
* [ ] vxlan.underlay
* [ ] vxlan.overlay
* [ ] vxlan.policy
* [ ] vxlan.multisite
* [ ] defaults.vxlan
* [ ] other

## Proposed Changes

change default value for edge bgp password enable.

## Checklist

### DIFF
--- a/roles/dtc/common/templates/ndfc_edge_connections.j2
+++ b/roles/dtc/common/templates/ndfc_edge_connections.j2
@@ -17,8 +17,12 @@
           description: {{ 'nace_bgp_peer_template_dci_underlay_jython_' + link.source_device + '_' + link.source_interface + '_' + link.dest_device }}
           name: bgp_peer_template_dci_underlay_jython
           policy_vars:
-            BGP_PASSWORD: "{{ link.bgp_section.bgp_password | default('') }}"
-            BGP_PASSWORD_ENABLE: {{ link.bgp_section.bgp_password_enable | default(defaults.vxlan.topology.edge_connections.bgp_section.bgp_password_enable) }}
+{% if link.bgp_section is defined and link.bgp_section.bgp_password is defined and link.bgp_section.bgp_password is not none and link.bgp_section.bgp_password | length > 0 %}
+            BGP_PASSWORD_ENABLE: true
+            BGP_PASSWORD: "{{ link.bgp_section.bgp_password }}"
+{% else %}
+            BGP_PASSWORD_ENABLE: false
+{% endif %}
             TEMPLATE_NAME: {{ data_model_extended.vxlan.fabric.name + '-' + link.dest_fabric + '-IPV4-EBGP' }}
             NEIGHBOR_ASN: "{{ link.bgp_section.neighbor_asn }}"
             OVERRIDE_LOCAL_ASN: false
--- a/roles/dtc/common/templates/ndfc_edge_connections.j2
+++ b/roles/dtc/common/templates/ndfc_edge_connections.j2
@@ -17,10 +17,11 @@
           description: {{ 'nace_bgp_peer_template_dci_underlay_jython_' + link.source_device + '_' + link.source_interface + '_' + link.dest_device }}
           name: bgp_peer_template_dci_underlay_jython
           policy_vars:
-{% if link.bgp_section is defined and link.bgp_section.bgp_password is defined and link.bgp_section.bgp_password is not none and link.bgp_section.bgp_password | length > 0 %}
-            BGP_PASSWORD_ENABLE: true
+{% if link.bgp_section.bgp_password_enable | default(defaults.vxlan.topology.edge_connections.bgp_section.bgp_password_enable) | bool %}
             BGP_PASSWORD: "{{ link.bgp_section.bgp_password }}"
+            BGP_PASSWORD_ENABLE: true
 {% else %}
+            BGP_PASSWORD: ""
             BGP_PASSWORD_ENABLE: false
 {% endif %}
             TEMPLATE_NAME: {{ data_model_extended.vxlan.fabric.name + '-' + link.dest_fabric + '-IPV4-EBGP' }}
--- a/roles/validate/files/defaults.yml
+++ b/roles/validate/files/defaults.yml
@@ -259,7 +259,7 @@ factory_defaults:
         source_interface_mtu: 9216
         source_interface_enabled: true
         bgp_section:
-          bgp_password_enable: true
+          bgp_password_enable: false
     underlay:
       general:
         routing_protocol: ospf
--- a/roles/validate/files/defaults.yml
+++ b/roles/validate/files/defaults.yml
@@ -259,7 +259,7 @@ factory_defaults:
         source_interface_mtu: 9216
         source_interface_enabled: true
         bgp_section:
-          bgp_password_enable: false
+          bgp_password_enable: true
     underlay:
       general:
         routing_protocol: ospf
--- a/roles/validate/files/rules/common/321_topology_edge_connections.py
+++ b/roles/validate/files/rules/common/321_topology_edge_connections.py
@@ -1,6 +1,6 @@
 class Rule:
     id = "321"
-    description = "Verify edge connections have valid source devices and interfaces"
+    description = "Verify edge connections have valid source devices, interfaces, and bgp password configuration"
     severity = "HIGH"
 
     @classmethod
@@ -40,6 +40,26 @@ class Rule:
             if source_interface and '.' in source_interface:
                 results.append(
                     f"vxlan.topology.edge_connections.{source_device} source_interface must not contain a '.' (sub-interfaces are not allowed)."
+                )
+
+            # Verify bgp_password is provided when bgp_password_enable is true
+            bgp_section = edge_connection.get("bgp_section", {})
+            defaults = data_model.get("defaults", {})
+            default_bgp_pw_enable = (
+                defaults.get("vxlan", {})
+                .get("topology", {})
+                .get("edge_connections", {})
+                .get("bgp_section", {})
+                .get("bgp_password_enable", False)
+            )
+            bgp_password_enable = bgp_section.get("bgp_password_enable", default_bgp_pw_enable)
+            bgp_password = bgp_section.get("bgp_password", "")
+
+            if bgp_password_enable and (not bgp_password or str(bgp_password).strip() == ""):
+                results.append(
+                    f"vxlan.topology.edge_connections.{source_device} "
+                    f"bgp_password_enable is true but bgp_password is not provided. "
+                    f"A non-empty bgp_password is required when bgp_password_enable is true."
                 )
 
         return results


mthurstocisco:
Hey Charlie, 
So, had a bit of a think, I wonder if something like the below makes more sense and we just dont use a default here? basically, if you provide a password, the jinja template sets password to true, if you dont provide a password (or its blank), set password to false?

      policies:
        - create_additional_policy: False
          description: {{ 'nace_bgp_peer_template_dci_underlay_jython_' + link.source_device + '_' + link.source_interface + '_' + link.dest_device }}
          name: bgp_peer_template_dci_underlay_jython
          policy_vars:
            BGP_PASSWORD: "{{ link.bgp_section.bgp_password | default('') }}"
{% if link.bgp_section is defined and link.bgp_section.bgp_password is defined and link.bgp_section.bgp_password is not none %}
            BGP_PASSWORD_ENABLE: true
{% else %}
            BGP_PASSWORD_ENABLE: false
{% endif %}

peter8498 (approved):
lgtm!

roles/dtc/common/templates/ndfc_edge_connections.j2
```diff
@@ -17,8 +17,12 @@
           description: {{ 'nace_bgp_peer_template_dci_underlay_jython_' + link.source_device + '_' + link.source_interface + '_' + link.dest_device }}
           name: bgp_peer_template_dci_underlay_jython
           policy_vars:
-            BGP_PASSWORD: "{{ link.bgp_section.bgp_password | default('') }}"
-            BGP_PASSWORD_ENABLE: {{ link.bgp_section.bgp_password_enable | default(defaults.vxlan.topology.edge_connections.bgp_section.bgp_password_enable) }}
+{% if link.bgp_section is defined and link.bgp_section.bgp_password is defined and link.bgp_section.bgp_password is not none and link.bgp_section.bgp_password | length > 0 %}
```
mikewiebe: ### Medium: Preserve the public `bgp_password_enable` contract

**Issue**

The changed template derives `BGP_PASSWORD_ENABLE` only from whether `bgp_password` is non-empty. It never reads the schema-valid `bgp_password_enable` input. That silently changes the public contract from an explicit enable control, defaulting to `true`, to password-presence inference.

**Evidence**

- The changed branch [sets `BGP_PASSWORD_ENABLE: true` whenever a non-empty password exists and `false` otherwise](https://github.com/netascode/ansible-dc-vxlan/pull/803/files#diff-93e887c6e3d3381556765ad816decb4a1cd4f43b19653c95fc2dd93c4703dc43R20-R24); `link.bgp_section.bgp_password_enable` is never consulted.
- At authoritative model ref `f997000b00b383725a9853c7f200b3a2c38cbb08`, [`bgp_password_enable` and `bgp_password` are independent optional properties](https://wwwin-github.cisco.com/netascode/nac-vxlan/blob/f997000b00b383725a9853c7f200b3a2c38cbb08/schemas/schema.yaml#L402-L405), while the [enable default remains `true`](https://wwwin-github.cisco.com/netascode/nac-vxlan/blob/f997000b00b383725a9853c7f200b3a2c38cbb08/defaults/defaults.yaml#L266-L270). The [published generated documentation](https://netascode.cisco.com/docs/data_models/vxlan/topology/topology_edge_connection/) exposes the same Boolean default and optional password.
- Focused validation against that exact schema accepted both contradictory combinations: password present plus `bgp_password_enable: false`, and password omitted plus `bgp_password_enable: true`.
- A strict render of the PR template produced these effective payloads:

  | Input combination | Rendered enable | Password key present |
  | --- | ---: | ---: |
  | Password omitted, enable omitted | `false` | No |
  | Password empty or null | `false` | No |
  | Password present, enable omitted | `true` | Yes |
  | Password present, explicit enable `false` | `true` | Yes |
  | Password omitted, explicit enable `true` | `false` | No |

**Practical example**

An operator can pre-stage a BGP password in the model while deliberately keeping authentication disabled until the remote peer is ready:

```yaml
vxlan:
  topology:
    edge_connections:
      - source_device: border-1
        source_interface: Ethernet1/1
        dest_device: wan-router-1
        dest_fabric: WAN
        local_ip: 192.0.2.1/30
        neighbor_ip: 192.0.2.2
        bgp_section:
          neighbor_asn: 65100

          # The password is staged, but authentication must remain disabled.
          bgp_password_enable: false
          bgp_password: "example-staged-secret"
```

The explicit intent is:

```yaml
BGP_PASSWORD_ENABLE: false
BGP_PASSWORD: "example-staged-secret"
```

PR 803 ignores the Boolean and sees only that the password is non-empty, so it renders:

```yaml
BGP_PASSWORD_ENABLE: true
BGP_PASSWORD: "example-staged-secret"
```

If the remote router has not enabled the same authentication yet, the NaC run unexpectedly enables it on `border-1`; the peers no longer agree and the established BGP session can drop and remain down. The inverse is also silent: `bgp_password_enable: true` with an accidentally omitted password renders `BGP_PASSWORD_ENABLE: false` instead of failing validation for incomplete input.

**Existing PR overlap**

No matching existing PR comment found. The August 6 suggestion from `mthurstocisco` proposes password-presence gating, but it does not reconcile, remove, or deprecate the existing public enable property.

**Impact**

A user who explicitly disables password authentication while retaining a password gets authentication enabled. That can reset an established BGP session and prevent it from returning if the peer is not using the same password. Conversely, explicit enable intent without a password is silently disabled instead of being rejected as incomplete input. In both cases, a schema-valid setting is accepted and then contradicted by the rendered intent.

**Suggested fix**

Choose one authoritative contract and enforce it end to end. If the Boolean remains public, honor its effective value; when true, require a non-empty password through semantic validation, and when false keep enable false even if a password is supplied. If passwordless edge BGP should be the normal omission behavior, coordinate a source-model default change from true to false rather than overriding the source default in this template. In either case, emit the desired password key deterministically as described in Finding 2. Add schema/semantic/render tests for omitted, empty, null, present, explicit false plus present, and explicit true plus absent.

If password presence is intentionally replacing the Boolean, coordinate a source-model change that deprecates or removes `bgp_password_enable`, updates its default and generated docs, rejects ambiguous legacy combinations with a migration message, and lands before this behavior is released.

mikewiebe:
### Medium: Emit an empty password so removal is observable to `dcnm_policy`

**Issue**

The false branch omits `BGP_PASSWORD` entirely. `dcnm_policy` compares only keys present in desired `nvPairs`; it ignores additional keys already stored by NDFC. An existing disabled policy containing a password can therefore match the disabled/no-password payload, preventing the password from being cleared.

**Evidence**

- The PR emits only [`BGP_PASSWORD_ENABLE: false`](https://github.com/netascode/ansible-dc-vxlan/pull/803/files#diff-93e887c6e3d3381556765ad816decb4a1cd4f43b19653c95fc2dd93c4703dc43R23-R24) when no password is supplied. The previous template always emitted `BGP_PASSWORD`, using an empty string when absent.
- In DCNM 3.12.1, [`dcnm_policy_compare_nvpairs()` iterates only desired keys](https://github.com/CiscoDevNet/ansible-dcnm/blob/6311b66ff02b8ed593cbbfed4abdd502607dd6d2/plugins/modules/dcnm_policy.py#L743-L753).
- An exact method test returned `DCNM_POLICY_MATCH` for desired disabled/no-password state versus existing disabled/non-empty-password state.
- This remains unchanged on current DCNM `develop` commit `939c75bb667498631e9047ff9f62f424f00b98c2`, which is planned for the next release.

**Practical example**

Assume an earlier run left this policy in NDFC:

```yaml
existing_in_ndfc:
  BGP_PASSWORD_ENABLE: false
  BGP_PASSWORD: "<redacted-existing-secret>"
```

The operator removes `bgp_password` from the NaC model because the secret must be cleared. PR #803 renders:

```yaml
desired_from_nac:
  BGP_PASSWORD_ENABLE: false
  # BGP_PASSWORD is absent
```

The comparator checks `BGP_PASSWORD_ENABLE`, sees `false == false`, never examines the extra stored password, and returns:

```text
DCNM_POLICY_MATCH
```

No update is sent, so NDFC retains the secret. Publishing current DCNM `develop` does not change this result.

The always-present assignment proposed by `mthurstocisco` is therefore important for reset semantics:

```yaml
desired_from_nac:
  BGP_PASSWORD_ENABLE: false
  BGP_PASSWORD: ""
```

Now the comparison sees empty versus non-empty and detects an update.

**Existing PR overlap**

This strengthens the [unanswered August 6 suggestion from `mthurstocisco`](https://github.com/netascode/ansible-dc-vxlan/pull/803#issuecomment-5204296719). The current implementation adopts its enable gating but drops its always-present `BGP_PASSWORD` assignment.

**Impact**

Removing a password from the source model can leave the secret stored in NDFC while the run reports no policy change. It could later reappear if authentication is re-enabled.

**Suggested fix**

Always emit `BGP_PASSWORD`, resolving absent or null values to `""`, and compute `BGP_PASSWORD_ENABLE` separately according to the public model contract. Add a transition test asserting that existing disabled/non-empty state changes to desired disabled/empty state and that NDFC clears the stored value.

ccoueffe:
Hi Mike,

Thanks for the thorough review. I went back to my initial approach — keeping the boolean — but with a hybrid logic:

- We honor `bgp_password_enable` as the public contract
- If only the password is defined (without explicit boolean), it works because the default is `true` — so it just pushes
- We now catch the case where enable is `true` but password is empty/missing with a validation error

This covers the points raised by Matt and Peter during the meeting, your review comments, and my original concern about the empty password failure.

**Template changes (`ndfc_edge_connections.j2`):**
- When `bgp_password_enable` is `true`: sends the password as-is
- When `bgp_password_enable` is `false`: sends `BGP_PASSWORD: ""` — aligned with ND behavior where disabling eBGP password authentication clears the password
- `BGP_PASSWORD` is **always emitted** so `dcnm_policy` can detect password removal (addresses your finding about `dcnm_policy_compare_nvpairs` only comparing desired keys)

**Validation rule (321):**
- If `bgp_password_enable` is `true` (explicit or default) and `bgp_password` is empty/missing → **HIGH severity validation error**
- This catches the original bug (default `true` + no password = NDFC deploy failure) at validation time instead of deploy time

**Regarding the staged password scenario** (`bgp_password_enable: false` + password provided):
- In ND, disabling eBGP password removes the password — there is no concept of "staged but disabled"
- Our template mirrors this: `false` = empty password, regardless of what is in the model
- Since NaC is declarative (full state pushed every run), re-enabling just requires setting `bgp_password_enable: true` + password in the model and re-running

**Behavior matrix:**

| `bgp_password_enable` | `bgp_password` | Result |
|---|---|---|
| true (explicit or default) | provided | `ENABLE: true, PW: "secret"` |
| true (explicit or default) | missing/empty | ❌ Validation error |
| false | provided | `ENABLE: false, PW: ""` |
| false | missing/empty | `ENABLE: false, PW: ""` |

peter8498 (approved):
Looks good

mthurstocisco:
Hi @ccoueffe my worry is that to transition away from the current setup and defaults (IE enable is true) then, at least for our customers, your change will silently turn off authentication with no warnings, I think that is too much of a breaking change. Currently we have customers just specifying a password (as that was all that was required previously), with your change (based on your matrix above) they will have their authentication turned off and password removed. Please can you look at this logic again to better handle this breaking change transition

PR #815, opened 2026-06-19
Fix securityGroupTagMacSegmentation version-gate threshold (ndfc_version >= 12.5.0)

### LINKED ISSUE #814: securityGroupTagMacSegmentation version gate uses ndfc_version instead of nd_version (always true)
### Summary

PR #812 added `securityGroupTagMacSegmentation` / `securityGroupTagMacSegmentation_PREV` to `roles/dtc/common/templates/ndfc_fabric/dc_vxlan_fabric/security/dc_vxlan_fabric_security.j2`, gated for ND 4.2.1+. The gate compares the wrong variable: it uses `ndfc_version` (NDFC scheme, e.g. `12.4.1.321`) against the ND-scheme string `4.2.1`.

Since every NDFC release is `12.x`, `version_compare` always evaluates `12.x >= 4.2.1` as true, so the field is sent to every controller — including ND 4.1, which does not support it.

### Impact

The result is a hard failure at fabric create/update (`dcnm_fabric`, state `merged`), not just at test time. It hits any VXLAN EVPN fabric with IPv4 underlay (`enable_ipv6_underlay: false`) running on a controller older than the ND 4.2.1 build.

Observed against ND 4.1.1g (NDFC 12.4.1.321):

```
fatal: [nac-fabric2-ca]: FAILED! ... dcnm_fabric ... RETURN_CODE 500
"DATA": "Invalid JSON response: Failed to create the fabric, due to invalid fields
[{securityGroupTagMacSegmentation=false, securityGroupTagMacSegmentation_PREV=false}]"
```

### Root cause

The codebase convention is that `ndfc_version` is compared against `12.x` numbers, while ND-version gating uses `nd_version` normalized with `regex_replace` (see `dc_vxlan_fabric_freeform.j2`). The new gate mixes the two: it reads `ndfc_version` but compares against an ND-scheme number.

### Scope

Currently only on `develop` (merged after the 0.8.0 release of 2026-05-31), so released versions are not affected — but this will ship broken to every client on NDFC older than the ND 4.2.1 build the moment a release is cut from `develop`. Should be fixed before the next release.

### Fix

Gate on `nd_version` following the existing convention:

```jinja2
{% if nd_version | regex_replace('^(\\d+\\.\\d+\\.\\d+).*', '\\1') | cisco.nac_dc_vxlan.version_compare('4.2.1', '>=') %}
```

dacasti2mx:
## Related Issue(s)

Fixes #814

## Related Collection Role

## Related Data Model Element

* [x] other (template-only version-gate fix — no data model change)

## Proposed Changes

#812 gates `securityGroupTagMacSegmentation` with `ndfc_version | version_compare('4.2.1', '>=')`. The variable is `ndfc_version` (NDFC scheme, e.g. `12.4.1.321`), but the threshold `4.2.1` is an ND-scheme number, so the comparison is always true (every NDFC release is `12.x`, and `12.x >= 4.2.1`). The field is therefore sent to every controller — including ND 4.1, which does not support it and rejects fabric create/update with HTTP 500.

This corrects the threshold to the NDFC version that ships with ND 4.2.1, keeping `ndfc_version` as-is:

- ND 4.1.1g → NDFC `12.4.1.321`
- ND 4.2.1 → NDFC `12.5.0.475`

so the gate becomes `ndfc_version | version_compare('12.5.0', '>=')`.

## Test Notes

Validated against Nexus Dashboard 4.1.1g (NDFC 12.4.1.321):

- Before: fabric create failed with `RETURN_CODE 500` — `invalid fields [{securityGroupTagMacSegmentation=false, securityGroupTagMacSegmentation_PREV=false}]`.
- After: the field is no longer emitted on ND 4.1, and fabric create/deploy completes successfully.

4.1.1g (NDFC 12.4.1.321); feature targets ND 4.2.1+ (NDFC 12.5.0.475).

## Checklist

* [ ] New or updates to documentation has been made accordingly (N/A — template-only fix)

### DIFF
--- a/roles/dtc/common/templates/ndfc_fabric/dc_vxlan_fabric/security/dc_vxlan_fabric_security.j2
+++ b/roles/dtc/common/templates/ndfc_fabric/dc_vxlan_fabric/security/dc_vxlan_fabric_security.j2
@@ -1,7 +1,7 @@
 {# Auto-generated NDFC DC VXLAN EVPN Security config data structure for fabric {{ vxlan.fabric.name }} #}
 {% if not (vxlan.underlay.general.enable_ipv6_underlay | default(defaults.vxlan.underlay.general.enable_ipv6_underlay) | ansible.builtin.bool) %}
   ENABLE_SGT: false
-{% if ndfc_version | cisco.nac_dc_vxlan.version_compare('4.2.1', '>=') %}
+{% if ndfc_version | cisco.nac_dc_vxlan.version_compare('12.5.0', '>=') %}
   securityGroupTagMacSegmentation: false
   securityGroupTagMacSegmentation_PREV: false
 {% endif %}


roles/dtc/common/templates/ndfc_fabric/dc_vxlan_fabric/security/dc_vxlan_fabric_security.j2
```diff
@@ -1,7 +1,7 @@
 {# Auto-generated NDFC DC VXLAN EVPN Security config data structure for fabric {{ vxlan.fabric.name }} #}
 {% if not (vxlan.underlay.general.enable_ipv6_underlay | default(defaults.vxlan.underlay.general.enable_ipv6_underlay) | ansible.builtin.bool) %}
   ENABLE_SGT: false
-{% if ndfc_version | cisco.nac_dc_vxlan.version_compare('4.2.1', '>=') %}
+{% if nd_version | regex_replace('^(\\d+\\.\\d+\\.\\d+).*', '\\1') | cisco.nac_dc_vxlan.version_compare('4.2.1', '>=') %}
```
mikewiebe: Good catch @dacasti2mx 

To be consistent with how we use this in other places in the templates shouldn't we just compare against `NDFC 12.4.1.321` instead?

dacasti2mx:
Good point on consistency — and it turns out the cleanest fix keeps `ndfc_version` exactly as in #812 and just corrects the threshold.

One catch on `12.4.1.321`: that's exactly the NDFC version that ships with **ND 4.1.1g** — the controller this fails on. With `>=`, `12.4.1.321 >= 12.4.1.321` is true, so the field would still be emitted on ND 4.1 and we'd hit the same 500.

The NDFC versions are ND 4.1.1g → `12.4.1.321` and ND 4.2.1 → `12.5.0.475` (the line jumps `12.4.1` → `12.5.0`; there is no `12.4.2`). So the right threshold is `12.5.0`. Updated the PR — the diff is now just `version_compare('4.2.1', '>=')` → `version_compare('12.5.0', '>=')`, keeping `ndfc_version`.

mikewiebe (approved):

PR #819, opened 2026-06-23
Fix #813: continue create pipeline on config-save HTTP 500

dacasti2mx:
## Related Issue(s)
Fixes #813

## Related Collection Role
* [ ] cisco.nac_dc_vxlan.validate
* [x] cisco.nac_dc_vxlan.dtc.create
* [ ] cisco.nac_dc_vxlan.dtc.deploy
* [ ] cisco.nac_dc_vxlan.dtc.remove
* [ ] other

## Related Data Model Element
* [x] vxlan.topology

## Proposed Changes
On a pre-provisioned fabric the switches can still be reloading or
unreachable when the create pipeline runs the intermediate config-save.
NDFC then returns HTTP 500. 0.7.2 logged this and continued; 0.8.0 aborts
at `inventory_config_save`, so the rest of the fabric intent is never
applied. This is a regression.

**Root cause:** `pipeline_base._config_save` already meant to treat a 500 as
non-fatal, but the check could never fire. `fabric_deploy_manager` discarded
the NDFC response on a failed config-save and returned `failed=True` with no
`msg`, so the `RETURN_CODE` was lost before the caller could look at it.

**Fix:**
- `fabric_config_save()` returns the NDFC response.
- `manage_fabrics()` propagates it into `results['msg']` when config-save
  fails, so the `RETURN_CODE` survives.
- `_config_save()` treats `RETURN_CODE == 500` as non-fatal and emits a clear
  warning (full detail under `-v`). Any other failure — 400, auth, or a
  malformed/missing response — still aborts the run. This is stricter and
  safer than the old 0.7.2 rescue, which swallowed every error.

Only the create pipeline registers `_config_save` (`create_resources.yml`);
`remove_resources.yml` does not, and the deploy role uses `operation: all`,
so deploy and remove behavior are unchanged.

## Test Notes
The collection has no unit-test harness today, and CI runs only build,
sanity, and lint — so a proper unit test needs new infrastructure. To avoid
blocking the 0.8.1 fix on that project-structure decision, tests are
deferred and tracked in a follow-up: #818.

Verified locally with a temporary mock harness (not committed) that drives
the real `_config_save`:
- HTTP 500 -> run continues, warning emitted
- HTTP 400 -> run aborts (fatal)
- missing or malformed `msg` -> run aborts (fatal)
- HTTP 200 -> success

Deferred unit-test matrix (in the follow-up):
- 500 -> manager returns `failed=True` with the response intact
- 500 -> `_config_save` returns non-fatal
[...]

PR #842, opened 2026-07-20
Handle transient ToR-vPC greenfield config-save failures

### LINKED ISSUE #840: [0.8.1] Greenfield ToR-vPC deployment aborts at intermediate vpc_config_save
### Ansible Version

ansible [core 2.16.3]
  python version = 3.11.15
  jinja version = 3.1.6

Collection          Version
------------------- --------
cisco.nac_dc_vxlan  0.8.1-dev (develop @ 11588f44)
cisco.dcnm          3.12.1-dev
cisco.nxos          10.2.0
ansible.netcommon   7.1.0
ansible.posix       2.0.0
ansible.utils       5.1.2
community.general   10.1.0

4.1.1g (NDFC 12.4.1.321)

### Cisco NX-OS Version

10.6(2), virtual N9K-C9300v

cisco.nac_dc_vxlan.dtc.create

vxlan.topology

### Expected Behavior

When a fabric declares both a parent-leaf vPC and a ToR-vPC pair together,
the create workflow should be able to finish the deployment. NDFC needs the
ToR-to-leaf association (created later, by tor_pairing) before it can fully
accept the ToR-vPC config — so an intermediate error at this point is
expected, and the workflow shouldn't abort permanently because of it.

### Actual Behavior

If vxlan.topology.vpc_peers declares both pairs in the same run, the create
pipeline always processes them in this order: vpc_peering, then
vpc_config_save, then tor_pairing, then tor_config_save.

vpc_peering succeeds first and creates both vPC objects in NDFC. But the
vpc_config_save step right after it fails — NDFC rejects it because the
ToR-to-leaf association doesn't exist yet. That association is only created
in the next step, tor_pairing, which never gets a chance to run:

fatal: [<fabric>]: FAILED! => {
  "msg": {
    "RETURN_CODE": 500,
    "DATA": "Invalid JSON response: Switch [<leaf>/<leaf-serial>]: Unsupported
             topology: tor cannot be vPC if the parent leaf is not vPC<br>
             [<leaf-serial>/<leaf>] - Unexpected error generating Tor
             configuration. Last Status [Tor Association] "
  }
}

The run stops here — tor_pairing and tor_config_save never get a chance to
run. `show feature | include vpc` on the switches afterward confirms
`feature vpc` was never actually pushed, even though vpc_peering already
succeeded and created both vPC objects in NDFC (changed=true, HTTP 200).

This scenario is related to #813 and #819 but is not covered by the
pre-provisioning-specific handling added by #819 — this is a from-scratch
greenfield deployment with no pre-provisioned switches involved.

### Ansible Playbook

---
- name: Manage VXLAN EVPN Fabric
  hosts: all
  gather_facts: false
  roles:

### Data Model

# Abbreviated data-model fragment
vxlan:
  topology:
    switches:
      - name: leaf1
        role: leaf
        serial_number: LEAF_SERIAL_1

      - name: leaf2
        role: leaf
        serial_number: LEAF_SERIAL_2

      - name: tor1
        role: tor
        serial_number: TOR_SERIAL_1

      - name: tor2
        role: tor
        serial_number: TOR_SERIAL_2

    vpc_peers:
      - peer1: leaf1
        peer2: leaf2
        peer1_peerlink_interfaces:
          - name: Ethernet1/11
          - name: Ethernet1/12
        peer2_peerlink_interfaces:
          - name: Ethernet1/11
          - name: Ethernet1/12

      - peer1: tor1
        peer2: tor2
        peer1_peerlink_interfaces:
          - name: Ethernet1/11
          - name: Ethernet1/12
        peer2_peerlink_interfaces:
          - name: Ethernet1/11
          - name: Ethernet1/12

    tor_peers:
      - tor1: tor1
        tor2: tor2
        parent_leaf1: leaf1
        parent_leaf2: leaf2

### Steps to Reproduce

1. Declare a greenfield VXLAN EVPN fabric where vxlan.topology.vpc_peers
   contains both the parent-leaf vPC pair and a ToR-vPC pair, and
   vxlan.topology.tor_peers declares the corresponding ToR-to-leaf pairing,
   all in the same initial run (nothing previously deployed).
2. Run cisco.nac_dc_vxlan.dtc.create against this fabric.
3. Observe vpc_peering (dcnm_vpc_pair) complete with changed=true.
4. Observe the following vpc_config_save (_config_save) step fail with
   RETURN_CODE 500 and the signature shown above.
5. Observe the run abort before tor_pairing or tor_config_save execute.

### Relevant Debug Output

CREATE [<fabric>] Step 9/21: vpc_peering (dcnm_vpc_pair) -> ok (changed=True)
CREATE [<fabric>] Step 10/21: vpc_config_save (_config_save) -> failed

dacasti2mx:
## Related Issue(s)
Fixes #840

## Related Collection Role
* [ ] other

## Related Data Model Element
* [ ] vxlan.fabric
* [ ] vxlan.global
* [x] vxlan.topology
* [ ] vxlan.underlay
* [ ] vxlan.overlay
* [ ] vxlan.policy
* [ ] vxlan.multisite
* [ ] defaults.vxlan
* [ ] other

## Proposed Changes

This adds a narrow, non-fatal exception for the transient HTTP 500 that
vpc_config_save can hit during ToR-vPC greenfield creation. It sits
alongside the existing pre-provisioned-switches exception from #819 —
that one is untouched and not broadened.

The bypass only kicks in when all of these are true:

1. It's the vpc_config_save step (inventory_config_save and
   tor_config_save are never eligible — they stay fatal).
2. The HTTP code is exactly 500.
3. tor_pairing and tor_config_save both survived --tags filtering for this
   run (if either was excluded, the deferred state could never actually get
   resolved, so the bypass doesn't apply).
4. changes_detected_tor_pairing is true for this run.
5. There's a pending vpc_to_vpc ToR pairing this run still needs to
   process:
   - in diff-run mode: at least one vpc_to_vpc entry in
     resource_data['tor_pairing']['diff']['updated'];
   - in full-run/discovery mode: at least one vpc_to_vpc entry in the
     data model, plus at least one leaf switch to discover against.
6. NDFC's error message contains the exact signature
   "tor cannot be vpc if the parent leaf is not vpc"
   (case/HTML/whitespace-tolerant).

Any other HTTP 500, any other status code, a missing/malformed response, or
a 500 with a different message — all of those stay fatal.

When the bypass fires, tor_config_save gets forced to run once, even if
tor_pairing itself reports changed=false. Normally tor_config_save only
runs when tor_pairing actually changed something (via
runtime_change_refs), but here the real config-save was deferred rather
than performed, so tor_config_save is the only step left that can actually
fix the state. If tor_pairing fails, the pipeline still stops immediately
before tor_config_save runs — that fail-fast behavior doesn't change.

This doesn't reorder create_resources.yml, and it doesn't touch
_fabric_has_preprovisioned_switches() or how it already behaves.

## Test Notes

I tested this with a local mock harness covering: the exact error signature
with a pending vpc_to_vpc pairing (non-fatal), a different signature
(fatal), no vpc_to_vpc pairing at all (fatal), the wrong step name (fatal),
the wrong HTTP code (fatal), a malformed response (fatal), the existing
pre-provisioned + 500 case (unchanged), a normal successful response
(unchanged), the pairing excluded via --tags (fatal),
changes_detected_tor_pairing being false (fatal), both diff-mode cases, and
the forced tor_config_save sequencing.

Results:

Config-save unit matrix: 14/14 PASS
run_pipeline integration matrix: 6/6 PASS

The test harness and captured results live in the integration repository,
not in this PR.

I also validated this live against NDFC 12.4.1.321 / Nexus Dashboard 4.1.1g
with virtual N9K-C9300v switches (NX-OS 10.6(2)): a greenfield ToR-vPC
deployment that used to abort at vpc_config_save now goes all the way
through tor_pairing and tor_config_save, feature vpc reaches every switch
involved, and running it again afterward (idempotent) doesn't re-trigger
the bypass.

4.1.1g (NDFC 12.4.1.321)

## Checklist

### DIFF
--- a/plugins/plugin_utils/pipeline_base.py
+++ b/plugins/plugin_utils/pipeline_base.py
@@ -47,6 +47,7 @@ __metaclass__ = type
 
 import json
 import os
+import re
 import time
 from abc import ABC, abstractmethod
 
@@ -94,6 +95,13 @@ class PipelineRunnerBase(ABC):
         self.executor = executor
         self.task_vars = task_vars
 
+        # Resource names that must bypass the runtime_change_refs guard (and
+        # only that guard -- data_model_guard, change_flag_guard and tag
+        # filtering still apply normally) for this run. Consumed one-shot via
+        # .discard() where checked. See _config_save() for the current sole
+        # producer.
+        self._forced_runtime_steps = set()
+
         # Load pipeline from registry
         collection_path = RegistryLoader.get_collection_path()
         registry = RegistryLoader.load(collection_path, self.REGISTRY_KEY)
@@ -132,6 +140,14 @@ class PipelineRunnerBase(ABC):
         pipeline = RegistryLoader.filter_pipeline_by_tags(
             pipeline, ansible_run_tags, self.role_tag
         )
+
+        # Cache the resource names that survived tag filtering for this run,
+        # so a step executing mid-pipeline (e.g. _config_save) can tell
+        # whether a *later* step was excluded by --tags/--skip-tags rather
+        # than assuming the full, unfiltered pipeline will run.
+        self._filtered_pipeline_resource_names = {
+            step['resource_name'] for step in pipeline
+        }
 
         # Hook: subclass pre-pipeline setup (e.g., pre-fetch switch list)
         context = self._pre_pipeline_setup()
@@ -218,6 +234,10 @@ class PipelineRunnerBase(ABC):
             # changed anything at runtime (i.e., NDFC module returned
             # changed=true).  This avoids unnecessary operations like
             # config-save when prior modules were idempotent.
+            #
+            # A resource listed in self._forced_runtime_steps bypasses this
+            # guard for one invocation, regardless of prior changed status.
+            # See _config_save() for why/when a step schedules this.
             runtime_refs = step.get('runtime_change_refs')
             if runtime_refs:
                 prior_changed = any(
@@ -225,7 +245,9 @@ class PipelineRunnerBase(ABC):
                     for sr in step_results
                     if sr.get('resource_name') in runtime_refs
                 )
-                if not prior_changed:
+                forced_run = resource_name in self._forced_runtime_steps
+                self._forced_runtime_steps.discard(resource_name)
+                if not prior_changed and not forced_run:
                     step_results.append({
                         'resource_name': resource_name,
                         'module': module,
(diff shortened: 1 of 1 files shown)


dacasti2mx:
Closing without merging: issue #840 was closed after follow-up testing failed to independently reproduce the bug. The fix itself remains sound and was validated live by the author; reopen if the failure resurfaces.

PR #847, opened 2026-07-21
Fix silently skipped deletion of the final MCFG VRFs/networks (unsafe sentinel serialization)

### LINKED ISSUE #846: [0.8.0/0.8.1] Removing the final MCFG VRFs or networks silently skips deletion
### Ansible Version

```
ansible [core 2.16.3]
  python version = 3.11.15
  jinja version = 3.1.6
```

```
Collection          Version
------------------- --------
cisco.nac_dc_vxlan  0.8.1-dev (develop @ 11588f44)
cisco.dcnm          3.12.1-dev
cisco.nxos          10.2.0
ansible.netcommon   7.1.0
ansible.posix       2.0.0
ansible.utils       5.1.2
community.general   10.1.0
```

4.1.1g (NDFC 12.4.1.321)

### Cisco NX-OS Version

10.6(2), virtual N9K-C9300v

`cisco.nac_dc_vxlan.dtc.remove`

`vxlan.multisite`

### Expected Behavior

When the final remaining `vxlan.multisite.overlay` VRF or network entries are removed and the corresponding list becomes empty for an MCFG fabric, the remove pipeline should detect the change and send the corresponding VRF/network delete requests to NDFC — the same way any other declared-then-removed resource is handled in diff-run mode.

### Actual Behavior

The collection keeps a small file on disk for each MCFG fabric (the **sentinel file**, `_msite_overlay_sentinel.yml`) to remember what the VRFs/networks looked like last time, so it can tell if anything changed.

The function that writes it — `_detect_msite_overlay_changes()` (`plugins/action/dtc/build_resource_data.py`) — doesn't save the data as plain values. Instead of writing `name: VRF01`, the file ends up with something like this for that same value:

```yaml
? !!python/object/apply:ansible.utils.unsafe_proxy.AnsibleUnsafeText
  - VRF01
```

That's a Python-specific instruction embedded in the file, not just plain text. The function that reads the file back — `_sentinel_had_key()`, same file — deliberately refuses to load files containing that kind of instruction (a normal security precaution, so it never rebuilds arbitrary code from a file). Because of this mismatch, reading the file always fails.

That failure is silently ignored instead of being reported. So instead of saying "I couldn't read my saved data," the collection just assumes "there was nothing here before."

This becomes a problem specifically once every VRF/network is removed: the current data doesn't disappear, it just becomes an empty list. The only way to tell "these existed before and were removed" from "there was never anything here" is by checking that saved file — and that check is exactly what's silently failing.

End result: the collection always assumes nothing existed before, so it thinks nothing changed, and it never asks NDFC to delete the VRF/networks. Since nothing fails loudly, the playbook finishes normally as if the removal worked — but the VRF/networks are still attached in NDFC.

- **API method:** none
- **API endpoint:** none
- **Controller requests:** 0

### Scope

**Affected:**
- MCFG fabrics where the final remaining `vxlan.multisite.overlay.vrfs` or `.networks` entries are removed, so that list becomes empty (e.g. removing the corresponding host_vars files or clearing the lists) in diff-run mode (`force_run_all: false`). Present since 0.8.0 (confirmed in the `0.8.0` tag's `plugins/action/dtc/build_resource_data.py`; the file did not exist prior to `0.7.2`).
- The failure occurs before any controller request and has no NDFC-version conditional — it is a client-side collection defect in how the collection serializes/deserializes its own local change-detection file. Confirmed on ND 4.1.1g / NDFC 12.4.1.321; expected on ND 4.2 as well, but ND 4.2 has not yet been tested directly.

**Not affected:**
- Removing some, but not all, entries from an otherwise still-populated vrfs/networks list (e.g. removing 2 of 3 networks while one remains) — that case is handled correctly via the normal per-resource diff mechanism, independent of this defect.
- Greenfield deployments (first-ever run for a given overlay) — there is no pre-existing sentinel to misread in that case.
- MSD fabrics share the same overlay change-detection code path but have not been separately reproduced/validated here — that remains open.

### Ansible Playbook

```yaml
---
- name: Manage VXLAN MCFG Fabric
  hosts: all
  gather_facts: false
  roles:
    - role: cisco.nac_dc_vxlan.dtc.remove
```

### Data Model

```yaml
# Previous desired state (already deployed and attached in NDFC)
vxlan:
  multisite:
    overlay:
      vrfs:
        - name: VRF01
          vlan_id: 3001
          vrf_attach_group: all
      networks:
        - name: Net01
          vrf_name: VRF01
          vlan_id: 3101
          net_id: 130101
          network_attach_group: all
      vrf_attach_groups:
        - name: all
          switches: [...]
      network_attach_groups:
        - name: all
          switches: [...]

# Current desired state: VRF01/Net01 removed. The extended/rendered data
# model keeps every overlay key present with an empty list rather than
# removing the "overlay" key itself -- this truthy-dict-of-empty-lists
# shape is what actually reaches the pipeline, and is the specific
# condition that triggers the defect:
vxlan:
  multisite:
    overlay:
      vrfs: []
      networks: []
      vrf_attach_groups: []
      network_attach_groups: []
```

### Steps to Reproduce

1. Deploy an MCFG parent fabric with `vxlan.multisite.overlay.vrfs`/`networks` declared and successfully created/attached in NDFC.
2. Remove every entry from `vxlan.multisite.overlay.vrfs` and `.networks` (or delete the corresponding data model files/host_vars entries), so both lists become empty.
3. Run the create → deploy → remove workflow in diff-run mode (`force_run_all: false`).
4. Observe the networks/vrfs remove steps report `skipped` with reason `change flag ... is False`, and the overall play reports success with no errors.
5. Query NDFC directly and confirm the VRFs/networks are still present and attached.

### Relevant Debug Output

```json
{
    "module": "dcnm_network",
    "reason": "change flag 'changes_detected_networks' is False",
    "resource_name": "networks",
    "status": "skipped"
},
{
    "module": "dcnm_vrf",
    "reason": "change flag 'changes_detected_vrfs' is False",
    "resource_name": "vrfs",
    "status": "skipped"
}
```

dacasti2mx:
## Related Issue(s)
Fixes #846

## Related Collection Role
- [ ] cisco.nac_dc_vxlan.validate
- [ ] cisco.nac_dc_vxlan.dtc.create
- [ ] cisco.nac_dc_vxlan.dtc.deploy
- [x] cisco.nac_dc_vxlan.dtc.remove
- [ ] other

## Related Data Model Element
- [ ] vxlan.fabric
- [ ] vxlan.global
- [ ] vxlan.topology
- [ ] vxlan.underlay
- [ ] vxlan.overlay
- [ ] vxlan.overlay_extensions
- [ ] vxlan.policy
- [x] vxlan.multisite
- [ ] defaults.vxlan
- [ ] other

## Proposed Changes

This replaces the sentinel read/write mismatch in `plugins/action/dtc/build_resource_data.py` that caused MCFG multisite overlay removal to be silently skipped.

**Root cause, in plain terms:** the collection keeps a small file per MCFG fabric (the sentinel, `_msite_overlay_sentinel.yml`) to remember what the overlay's VRFs/networks looked like last time, so the next run can tell if anything changed. The function that writes it — `_detect_msite_overlay_changes()` — didn't save plain values. Instead of `name: VRF01`, the file ended up with something like this for that same value:

```yaml
? !!python/object/apply:ansible.utils.unsafe_proxy.AnsibleUnsafeText
  - VRF01
```

That's a Python-specific instruction embedded in the file, not just plain text. The function that reads it back — `_sentinel_had_key()` — deliberately refuses to load a file containing that kind of instruction (a normal security precaution, so it never rebuilds arbitrary code from a file). The read always failed, and that failure was silently swallowed instead of surfaced, so the collection assumed "nothing existed before" every time.

That specific gap only matters once every VRF/network is removed: the current state becomes just an empty list either way, so the previous sentinel is the only place left to tell "these existed and were removed" from "there was never anything here." Since that read always silently failed, `changes_detected_vrfs`/`changes_detected_networks` never got set, `dcnm_network`/`dcnm_vrf` were never called, and nothing ever reached NDFC. Present since 0.8.0 (see #846).

**Fix** (single file, `plugins/action/dtc/build_resource_data.py`):

- Write with `yaml.safe_dump()` on values normalized to plain Python types (`_normalize_for_yaml()`), which raises `TypeError` rather than silently guessing via `str()` for any type it doesn't explicitly handle.
- Read with a single, load-before-mutate `_load_overlay_sentinel()` call, using `yaml.safe_load()` only — no custom loader. This read happens before the sentinel file (and the unrelated deferred-build cache file) are backed up, removed, or overwritten, so a bad read never leaves things half-touched.
- A sentinel this cannot read as a plain YAML mapping — the pre-existing Python-object-tagged format, a wrong top-level structure, or genuinely malformed YAML — now raises a `RuntimeError` naming the exact recovery step: re-run once with `force_run_all: true` and the complete, unchanged data model. A separate message is used for filesystem/permission failures, since `force_run_all: true` would not fix those.
- The failure-swallowing `_sentinel_had_key()` helper and its repeated sentinel reads are removed; the already-loaded sentinel is reused directly, so the sentinel is opened and parsed once per build instead of twice.

**Migration required for existing installations:** after upgrading, run once with `force_run_all: true` using the complete, unchanged data model before removing any VRFs/networks, so the existing sentinel gets regenerated in the new, plain-YAML format. Without this step, the next diff-run against a pre-upgrade sentinel will fail loud (by design) rather than silently skip the removal as before.

**Scope boundaries:**

- The reproduced and end-to-end validated scenario is MCFG. The modified sentinel implementation (`_detect_msite_overlay_changes()`, `_load_overlay_sentinel()`, `_write_overlay_sentinel()`) is shared by MCFG and MSD (`self.fabric_type in ('MSD', 'MCFG')`). No end-to-end MSD support claim is made by this PR — MSD has not been separately reproduced or validated.
- `_msite_build_overlay()` (`plugins/plugin_utils/pipeline_base.py`) is not modified by this PR. This PR covers the populated → empty transition. A separate, independent defect in the empty → populated (recreate) direction was found during this PR's live validation and is not fixed here — tracked separately, out of scope of this PR.

## Test Notes

**Local:** a 21-case/25-assertion matrix (`ResourceDataBuilder`/`ResourceRemover`/`PipelineRunnerBase`, no mocks of the fix itself) covering: the legacy sentinel format failing loud in diff-run (sentinel and cache file left byte-for-byte untouched); safe read/write round-tripping; missing/corrupt/wrong-structure/permission-denied sentinels (each producing the correct, distinct recovery message); full integration through a real `ResourceRemover.run_pipeline()` (controller-diff discovery, dispatch order, dispatch content); the run_map retry/self-healing path (unrelated to and unmodified by this fix); idempotent re-runs; and the full mandatory migration flow driven against the real, unmodified `ResourceDataBuilder.build()` (`force_run_all=True` → real `_cleanup_files()` → fresh safe sentinel → subsequent diff-run → real dispatch). Also covers several related scenarios to characterize `remove_resources.py`'s existing, unmodified behavior: only one resource type emptied while the other remains, a true partial removal within one resource type, and an attach-group-only change.

**Results:**

- Full matrix: **25/25 PASS**
- 19/25 assertions fail or differ against the baseline (reverted, original pre-fix code). Cases 20 and 21 pass on both versions because they characterize behavior unaffected by the sentinel defect (a real `diff.removed`-based partial removal, and an attach-group-only change that never needed the sentinel read to begin with), not evidence of the fix itself.

The test harness and captured results live in the integration repository, not in this PR.

**Live validation:** I also validated this live against ND 4.1.1g / NDFC 12.4.1.321 (virtual N9K-C9300v, NX-OS 10.6(2)), reproducing the exact reported scenario end to end. With the pre-migration (legacy) sentinel still on disk and a real data-model change (removing VRFs/networks) in diff-run mode, the build now fails with the exact `force_run_all: true` recovery message, leaving the sentinel completely untouched and zero controller requests made. `force_run_all: true` with the complete, unchanged data model then completes successfully and regenerates a plain, safe sentinel. The following diff-run correctly dispatches `dcnm_network`/`dcnm_vrf` with `state: deleted` (controller diff: 2 networks and 1 VRF on the controller, 0 in the data model, all discovered for deletion), and NDFC confirmed the actual deletion afterward.

Confirmed on 4.1.1g (NDFC 12.4.1.321). The failure occurs before any controller request and has no NDFC-version conditional in the code, so it is expected on ND 4.2 as well — ND 4.2 has not yet been tested directly.

## Checklist

- [x] Latest commit is rebased from develop with merge conflicts resolved
- [ ] New or updates to documentation has been made accordingly
- [ ] Assigned the proper reviewers

### DIFF
--- a/plugins/action/dtc/build_resource_data.py
+++ b/plugins/action/dtc/build_resource_data.py
@@ -116,6 +116,32 @@ class ResourceDataBuilder:
             return value.lower() in ('1', 'true', 'yes', 'on')
         return bool(value)
 
+    @staticmethod
+    def _normalize_for_yaml(obj):
+        """
+        Recursively convert Ansible-wrapped values (e.g. AnsibleUnsafeText)
+        to plain Python types so the result can be safely round-tripped
+        with yaml.safe_dump()/yaml.safe_load().
+
+        str.__str__() is used instead of plain str() because Ansible's
+        "unsafe" string wrapper overrides __str__ (and similar dunder
+        methods) to keep re-wrapping its own output as unsafe -- calling
+        the base str implementation directly is the reliable way to get a
+        genuinely plain str back.
+        """
+        if isinstance(obj, dict):
+            return {
+                ResourceDataBuilder._normalize_for_yaml(k): ResourceDataBuilder._normalize_for_yaml(v)
+                for k, v in obj.items()
+            }
+        if isinstance(obj, (list, tuple)):
+            return [ResourceDataBuilder._normalize_for_yaml(v) for v in obj]
+        if isinstance(obj, bool) or obj is None or isinstance(obj, (int, float)):
+            return obj
+        if isinstance(obj, str):
+            return str.__str__(obj)
+        raise TypeError(f"Unsupported sentinel value type: {type(obj).__name__}")
+
     def _should_run_structural_diff(self, diff_compare):
         """Structural diff data is only consumed during targeted diff runs."""
         return bool(diff_compare) and self.run_map_diff_run and not self.force_run_all
@@ -461,6 +487,17 @@ class ResourceDataBuilder:
         data model (including vrf_attach_groups and network_attach_groups)
         to a sentinel file and compares against the previous version.
 
+        The previous sentinel is loaded exactly once, via
+        _load_overlay_sentinel(), before ANY local state is touched --
+        the deferred-build cache file included. If that load fails
+        (sentinel written by a version of this collection before this
+        fix, or genuinely corrupted), it raises RuntimeError and this
+        method makes no further changes: the existing sentinel and cache
+        file are both left exactly as they were, no change flag is set,
+        and no create/remove step downstream can act on a value we
+        failed to read correctly. See _load_overlay_sentinel() for the
+        required recovery step.
+
         When a change is detected, sets BOTH the aggregate flag
         (changes_detected_msite_overlay) AND per-resource flags
         (changes_detected_vrfs, changes_detected_networks) so that
@@ -469,12 +506,6 @@ class ResourceDataBuilder:
         Also cleans up the deferred build cache file from any prior
         pipeline run in this playbook execution.
         """
-        # Clean up deferred build cache from prior pipeline runs so that
-        # each playbook run starts with a fresh cache.
-        cache_file = os.path.join(self.output_path, '_msite_overlay_cache.json')
-        if os.path.exists(cache_file):
-            os.remove(cache_file)
-
         overlay = (
             self.data_model
             .get('vxlan', {})
@@ -485,43 +516,46 @@ class ResourceDataBuilder:
         sentinel_file = os.path.join(self.output_path, '_msite_overlay_sentinel.yml')
         old_sentinel = sentinel_file + '.old'
 
-        # Handle empty/missing overlay: still need to detect removal of
-        # previously existing overlay data (user removed all VRFs/networks).
-        if not overlay:
-            if not os.path.exists(sentinel_file):
-                # No previous sentinel and no current overlay — nothing to detect
-                return
-            # Previous sentinel exists but overlay is now empty — detect removal
-            shutil.copy2(sentinel_file, old_sentinel)
-            os.remove(sentinel_file)
-            with open(sentinel_file, 'w') as f:
-                f.write(yaml.dump({}, default_flow_style=False, sort_keys=True))
-            if self._run_diff_model_changes(old_sentinel, sentinel_file):
-                if self.check_roles.get('save_previous', False):
-                    self.change_flags['changes_detected_msite_overlay'] = True
-                    self.change_flags['changes_detected_vrfs'] = True
-                    self.change_flags['changes_detected_networks'] = True
-                    display.v(
-                        f"COMMON [{self.fabric_name}] Multisite overlay "
-                        f"removed — all overlay data cleared"
-                    )
+        # Read-before-write: load whatever sentinel already exists on disk
+        # before this method makes any change to local state. A bad
+        # sentinel fails here, before the cache cleanup below or anything
+        # else runs, so a failed build never leaves things half-touched.
+        previous_overlay = self._load_overlay_sentinel(sentinel_file)
+        previous_had_vrfs = bool(previous_overlay.get('vrfs'))
+        previous_had_networks = bool(previous_overlay.get('networks'))
+
+        # Only now, after the sentinel is confirmed readable, clean up the
+        # deferred build cache from prior pipeline runs so that each
+        # playbook run starts with a fresh cache.
+        cache_file = os.path.join(self.output_path, '_msite_overlay_cache.json')
+        if os.path.exists(cache_file):
+            os.remove(cache_file)
+
+        # Nothing before, nothing now -- no sentinel to write, no change
+        # to detect.
+        if not overlay and not previous_overlay:
             return
 
-        # Backup previous sentinel file
-        if os.path.exists(sentinel_file):
-            shutil.copy2(sentinel_file, old_sentinel)
-            os.remove(sentinel_file)
+        # Current overlay is empty but a previous sentinel had content:
+        # the user removed all VRFs/networks from the data model. Record
+        # an empty sentinel and flag both resources for removal.
+        if not overlay:
+            self._write_overlay_sentinel(sentinel_file, old_sentinel, {})
+            if self.check_roles.get('save_previous', False):
+                self.change_flags['changes_detected_msite_overlay'] = True
+                self.change_flags['changes_detected_vrfs'] = True
+                self.change_flags['changes_detected_networks'] = True
+                display.v(
+                    f"COMMON [{self.fabric_name}] Multisite overlay "
+                    f"removed — all overlay data cleared"
+                )
+            return
 
         # Capture the ENTIRE overlay dict (vrfs, networks, vrf_attach_groups,
-        # network_attach_groups, etc.) for change detection. Previously only
-        # vrfs and networks were captured, missing attach group modifications.
-        overlay_content = yaml.dump(
-            overlay,
-            default_flow_style=False,
-            sort_keys=True,
-        )
-        with open(sentinel_file, 'w') as f:
-            f.write(overlay_content)
+        # network_attach_groups, etc.) for change detection -- not just
+        # vrfs and networks -- so attach-group-only edits are also caught.
+        normalized_overlay = self._normalize_for_yaml(overlay)
+        self._write_overlay_sentinel(sentinel_file, old_sentinel, normalized_overlay)
 
         overlay_vrfs = overlay.get('vrfs', [])
         overlay_networks = overlay.get('networks', [])
@@ -530,12 +564,12 @@ class ResourceDataBuilder:
         if self._run_diff_model_changes(old_sentinel, sentinel_file):
             if self.check_roles.get('save_previous', False):
                 self.change_flags['changes_detected_msite_overlay'] = True
-                # Set per-resource flags so pipeline steps pass their
-                # change_flag_guard checks. Both CREATE and REMOVE pipelines
-                # gate VRF/network steps on these flags.
-                if overlay_vrfs or self._sentinel_had_key(old_sentinel, 'vrfs'):
+                # A resource flag is set if it is non-empty now, or was
+                # non-empty before -- either way there is work to do:
+                # new items to create, or old ones to remove.
+                if overlay_vrfs or previous_had_vrfs:
                     self.change_flags['changes_detected_vrfs'] = True
-                if overlay_networks or self._sentinel_had_key(old_sentinel, 'networks'):
+                if overlay_networks or previous_had_networks:
                     self.change_flags['changes_detected_networks'] = True
                 display.v(
                     f"COMMON [{self.fabric_name}] Multisite overlay data "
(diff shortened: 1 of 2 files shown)


marehler (changes requested):
This fix does not work yet for MSD fabrics. When removing all vxlan.multisite.overlay config, changes are not detected. It would be great if you could extend the scope of this PR to cover MSD too.

The following change in plugins/action/dtc/remove_resources.py, def _resolve_step_data, line 204 seems to fix it for MSD fabrics too:
````
                and self.fabric_type in ('MCFG', 'MSD')
````

I observe one more issue. Task "REMOVE [VXLAN_MSD] Step 4/5: vrfs (dcnm_vrf)" fails with error message "Timeout waiting for VRF attachments." However, this seems to be a separate issue not related to this PR. Hence I suggest to handle it separately.

**Comment by juburnet** [comment]
@marehler  - I updated with your requested change to manage MSD.  

Can you re-review and also check the following condition in the remove_resources.py:

The _controller_diff_path method currently returns the onemanage API path only for MCFG, and falls back to the standard lan-fabric path for everything else. You need to verify whether MSD uses the same NDFC API path as MCFG or the standard one:
```
def _controller_diff_path(self, resource_name):
    if self.fabric_type == 'MCFG':          # ← does MSD need this too?
        return (
            "/onemanage/appcenter/cisco/ndfc/api/v1/onemanage/top-down"
            f"/fabrics/{self.fabric_name}/{resource_name}"
        )
    return (
        "/appcenter/cisco/ndfc/api/v1/lan-fabric/rest/top-down/v2"
        f"/fabrics/{self.fabric_name}/{resource_name}"
    )
If MSD also uses the onemanage top-down endpoint, this also becomes:
    if self.fabric_type in ('MCFG', 'MSD'):
```

marehler:
@juburnet, re. the_controller_diff_path: No, MSD uses the standard lan-fabric top-down path, which is what the current _controller_diff_path else branch returns. The onemanage endpoint is MCFG-specific (it's the newer one-manage fabric-group API that MCFG parents require, MSD parents use the classic lan-fabric API). For _resolve_step_data, both MSD and MCFG needed the empty-overlay fallback (same behavior). For _controller_diff_path, MSD and MCFG need different endpoints, i.e. the == 'MCFG' check must stay as-is.

marehler:
Tested with ND 4.3.1.

As mentioned in the description of the fix above, "the pre-existing Python-object-tagged format, a wrong top-level structure, or genuinely malformed YAML now raises a RuntimeError naming the exact recovery step: re-run once with force_run_all: true":

````
TASK [cisco.nac_dc_vxlan.common : Build Resources] **************************************************************************************************
COMMON [VXLAN_MSD] Built fabric: items=1, changed=False
COMMON [VXLAN_MSD] Built bgw_anycast_vip: items=0, changed=False
fatal: [vND43_MSD]: FAILED! => {"changed": false, "msg": "Build resource data failed: The existing multisite overlay sentinel is incompatible or corrupted. After upgrading the collection, run once with force_run_all: true using the complete, unchanged data model before removing VRFs or networks. Sentinel: /Users/marehler/Git/nac-vxlan-dev/nac-ndfc/collections/ansible_collections/cisco/nac_dc_vxlan/roles/dtc/common/files/msd/VXLAN_MSD/_msite_overlay_sentinel.yml. Error: could not determine a constructor for the tag 'tag:yaml.org,2002:python/object/apply:ansible.utils.unsafe_proxy.AnsibleUnsafeText'\n  in \"/Users/marehler/Git/nac-vxlan-dev/nac-ndfc/collections/ansible_collections/cisco/nac_dc_vxlan/roles/dtc/common/files/msd/VXLAN_MSD/_msite_overlay_sentinel.yml\", line 1, column 3"}
````
I suggest to note this in the changelog too.

For MSD, if you remove all vxlan.multisite.overlay config data, the networks/VRFs to be removed are now detected correctly. However, in my setup VRF removal fails. This seems to be a lab issue, not related to the changes of this PR, but I am not sure.

````
TASK [cisco.nac_dc_vxlan.remove : Execute Remove Resources] *****************************************************************************************

────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
REMOVE [VXLAN_MSD] Step 1/5: prepare_msite_data (_prepare_msite_data)
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
REMOVE [VXLAN_MSD] prepare_msite_data → ok (changed=False)REMOVE [VXLAN_MSD] prepare_msite_data → ok (changed=False) [5.1s]

────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
REMOVE [VXLAN_MSD] Step 2/5: msite_build_overlay (_msite_build_overlay)
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
REMOVE [VXLAN_MSD] msite_build_overlay → ok (changed=False) [0.0s]

────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
REMOVE [VXLAN_MSD] Step 3/5: networks (dcnm_network)
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
REMOVE [VXLAN_MSD] Controller diff for networks: 2 on controller, 0 in data model, 2 to delete
REMOVE [VXLAN_MSD] Executing dcnm_network for networks with state=deleted, save=None, deploy=None, items=2
REMOVE [VXLAN_MSD] networks → ok (changed=True) [22.4s]

────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
REMOVE [VXLAN_MSD] Step 4/5: vrfs (dcnm_vrf)
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
REMOVE [VXLAN_MSD] Controller diff for VRFs: 1 on controller, 0 in data model, 1 to delete
REMOVE [VXLAN_MSD] Executing dcnm_vrf for vrfs with state=deleted, save=None, deploy=None, items=1
REMOVE [VXLAN_MSD] vrfs → failed [545.4s]
fatal: [vND43_MSD]: FAILED! => {"changed": false, "msg": "Remove pipeline failed at step 'vrfs' (dcnm_vrf): Timeout waiting for VRF attachments. Pending: ['GREEN']", <...>

marehler (commented):
Could you please check if you observe the VRF removal failure too, or if it works fine in your setup?

dacasti2mx:
> Could you please check if you observe the VRF removal failure too, or if it works fine in your setup?

I tested PR #847 on NDFC 12.6.0.267 with `cisco.dcnm` 3.12.1. The two networks and the VRF were deleted successfully in about 100 seconds. The attachment sequence was:

`DEPLOYED -> PENDING -> OUT-OF-SYNC -> NA -> deleted`

I could not reproduce the timeout. It appears separate from PR #847 and may depend on the lab state.

dacasti2mx:
> Tested with ND 4.3.1.
> 
> As mentioned in the description of the fix above, "the pre-existing Python-object-tagged format, a wrong top-level structure, or genuinely malformed YAML now raises a RuntimeError naming the exact recovery step: re-run once with force_run_all: true":
> 
> ```
> TASK [cisco.nac_dc_vxlan.common : Build Resources] **************************************************************************************************
> COMMON [VXLAN_MSD] Built fabric: items=1, changed=False
> COMMON [VXLAN_MSD] Built bgw_anycast_vip: items=0, changed=False
> fatal: [vND43_MSD]: FAILED! => {"changed": false, "msg": "Build resource data failed: The existing multisite overlay sentinel is incompatible or corrupted. After upgrading the collection, run once with force_run_all: true using the complete, unchanged data model before removing VRFs or networks. Sentinel: /Users/marehler/Git/nac-vxlan-dev/nac-ndfc/collections/ansible_collections/cisco/nac_dc_vxlan/roles/dtc/common/files/msd/VXLAN_MSD/_msite_overlay_sentinel.yml. Error: could not determine a constructor for the tag 'tag:yaml.org,2002:python/object/apply:ansible.utils.unsafe_proxy.AnsibleUnsafeText'\n  in \"/Users/marehler/Git/nac-vxlan-dev/nac-ndfc/collections/ansible_collections/cisco/nac_dc_vxlan/roles/dtc/common/files/msd/VXLAN_MSD/_msite_overlay_sentinel.yml\", line 1, column 3"}
> ```
> 
> I suggest to note this in the changelog too.
> 
> For MSD, if you remove all vxlan.multisite.overlay config data, the networks/VRFs to be removed are now detected correctly. However, in my setup VRF removal fails. This seems to be a lab issue, not related to the changes of this PR, but I am not sure.
> 
> ```
> TASK [cisco.nac_dc_vxlan.remove : Execute Remove Resources] *****************************************************************************************
> 
> ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
> REMOVE [VXLAN_MSD] Step 1/5: prepare_msite_data (_prepare_msite_data)
> ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
> REMOVE [VXLAN_MSD] prepare_msite_data → ok (changed=False)REMOVE [VXLAN_MSD] prepare_msite_data → ok (changed=False) [5.1s]
> 
> ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
> REMOVE [VXLAN_MSD] Step 2/5: msite_build_overlay (_msite_build_overlay)
> ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
> REMOVE [VXLAN_MSD] msite_build_overlay → ok (changed=False) [0.0s]
> 
> ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
> REMOVE [VXLAN_MSD] Step 3/5: networks (dcnm_network)
> ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
> REMOVE [VXLAN_MSD] Controller diff for networks: 2 on controller, 0 in data model, 2 to delete
> REMOVE [VXLAN_MSD] Executing dcnm_network for networks with state=deleted, save=None, deploy=None, items=2
> REMOVE [VXLAN_MSD] networks → ok (changed=True) [22.4s]
> 
> ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
> REMOVE [VXLAN_MSD] Step 4/5: vrfs (dcnm_vrf)
> ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
> REMOVE [VXLAN_MSD] Controller diff for VRFs: 1 on controller, 0 in data model, 1 to delete
> REMOVE [VXLAN_MSD] Executing dcnm_vrf for vrfs with state=deleted, save=None, deploy=None, items=1
> REMOVE [VXLAN_MSD] vrfs → failed [545.4s]
> fatal: [vND43_MSD]: FAILED! => {"changed": false, "msg": "Remove pipeline failed at step 'vrfs' (dcnm_vrf): Timeout waiting for VRF attachments. Pending: ['GREEN']", <...>
> ```

I agree. 

Note for release prep: besides Fixes #846 in the Fixed list, 0.9.0 needs a Modified entry for the sentinel format change — an existing sentinel in the pre-fix !!python/object/apply: format is now rejected, so after upgrading a user must run once with force_run_all: true and the complete data model before removing any VRF or network from an MCFG or MSD fabric. Otherwise the first removal attempt after upgrade fails with an unexplained incompatible-sentinel error.

marehler (approved):
Ok, if you cannot reproduce the timeout and it works in your setup, I assume that's a lab issue not caused by the changes from this PR.

PR #855, opened 2026-07-24
Adding support for storm control on L2 interfaces and copy description on PC

### LINKED ISSUE #850: Enable/Disable CDP on L2 interfaces

cisco.nac_dc_vxlan.dtc.create

### New feature description:

```shell
Enable/Disable CDP on L2 interfaces
```

```shell

```

vxlan.topology

```shell
new key enable_cdp for L2 interfaces
```

```shell

```

```yaml
.
```

### Ansible Version

```shell

```

```shell

```

```shell

```

### Cisco NX-OS Version

```shell

```

### LINKED ISSUE #853: Enable Storm control on L2 interfaces

cisco.nac_dc_vxlan.dtc.create

### New feature description:

```shell
Enable Storm control on L2 interfaces
```

```shell
No
```

vxlan.topology

```shell

```

```shell

```

```yaml
.
```

### Ansible Version

```shell

```

```shell

```

```shell

```

### Cisco NX-OS Version

```shell

```

### LINKED ISSUE #854: access vPC defaults don't take effect
### Ansible Version

```shell
ansible [core 2.18.15]
```

```shell
Collection         Version   
------------------ ----------
ansible.netcommon  7.1.0     
ansible.posix      2.0.0     
ansible.utils      5.1.2     
cisco.dcnm         3.12.1-dev
cisco.nac_dc_vxlan 0.8.1-dev 
community.general  10.1.0
```

```shell
Akk
```

### Cisco NX-OS Version

```shell

```

cisco.nac_dc_vxlan.validate

vxlan.topology

### Expected Behavior

```shell
defaults for access_vpc are used when not all keys are present on the data model
```

### Actual Behavior

```shell
defaults for access_vpc are not used merged when not all keys are present on the data model, the trunk_vpc keys are used instead
```

### Ansible Playbook

```yaml
nsible-playbook -i inventory.yaml vxlan.yaml -l fabric
```

### Data Model

```yaml
NA
```

### Steps to Reproduce

```shell
run the playbook and check the values
```

### Relevant Debug Output

```shell

```

### LINKED ISSUE #857: vPC parameters are not checked when an existing vPC ID is present in the model, (even if the vPC id is in different pair)
### Ansible Version

```shell
ansible [core 2.18.15]
```

```shell
Collection         Version
------------------ ----------
ansible.netcommon  7.1.0
ansible.posix      2.0.0
ansible.utils      5.1.2
cisco.dcnm         3.12.1-dev
cisco.nac_dc_vxlan 0.8.1-dev
community.general  10.1.0
```

```shell
4.1
```

### Cisco NX-OS Version

```shell

```

cisco.nac_dc_vxlan.validate

vxlan.fabric

### Expected Behavior

```shell
vpc is validated correctly
```

### Actual Behavior

```shell
When multiple vpc id are defined and they are repeated, the second pair of interfaces is not review to check that the parameters are the same across the member interfaces.

Assume that fabric1-l1, fabric1-l2 defined a vpc interfaces with *vpc_id: 11*. The data model shown below doesn't fail even if the parameters are incorrect. In this case storm_control_unicast_level_percent.
```

### Ansible Playbook

```yaml
ansible-playbook -i inventory.yaml vxlan.yaml -l juarocha-nac-fabric1 --tags=role_validate
```

### Data Model

```yaml
---
vxlan:
  topology:
    switches:
      - name: fabric1-bl1
        interfaces:
          - name: port-channel11
            mode: trunk
            members:
              - eth1/8
            enable_storm_control: true
            storm_control_action: default
            storm_control_unicast_level_percent: 90.41412    <<<<<

      - name: fabric1-bl2
        interfaces:
          - name: port-channel11
            mode: trunk
            members:
              - eth1/8
            enable_storm_control: true
            storm_control_action: default
            storm_control_unicast_level_percent: 90.4    <<<<<
```

### Steps to Reproduce

```shell
Run the playbook
```

### Relevant Debug Output

```shell

```

juarocha:
## Related Issue(s)

Fixes #850
Fixes #853 
Fixes #854 
Fixes #857

## Related Collection Role

* [X] cisco.nac_dc_vxlan.dtc.create
* [ ] other

## Related Data Model Element

* [ ] vxlan.fabric
* [ ] vxlan.global
* [X] vxlan.topology
* [ ] vxlan.underlay
* [ ] vxlan.overlay
* [ ] vxlan.policy
* [ ] vxlan.multisite
* [ ] defaults.vxlan
* [ ] other

## Proposed Changes

For storm control, adding new keys for  under L2 interfaces:
- enable_storm_control
- storm_control_action
- storm_control_broadcast_level_percent
- storm_control_multicast_level_percent
- storm_control_unicast_level_percent
- storm_control_broadcast_level_pps
- storm_control_multicast_level_pps
- storm_control_unicast_level_pps

For copy description, adding new keys for under port-channel interfaces:
- copy_description

4.1

## Checklist

* [X] Latest commit is rebased from develop with merge conflicts resolved
* [X] New or updates to documentation has been made accordingly
* [X] Assigned the proper reviewers

### DIFF
--- a/roles/dtc/common/templates/ndfc_interfaces/ndfc_interface_trunk_po.j2
+++ b/roles/dtc/common/templates/ndfc_interfaces/ndfc_interface_trunk_po.j2
@@ -24,6 +24,7 @@
     admin_state: {{ interface['enabled'] | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.enabled) | lower }}
     mode: 'trunk'
     description: "{{ interface['description'] | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.description) }}"
+    copy_description: {{ interface['copy_description'] | default(defaults.vxlan.topology.switches.interfaces.topology_switch_access_po_interface.copy_description) | lower }}
     mtu: {{ interface['mtu'] | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.mtu) }}
     speed: {{ interface['speed'] | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.speed) }}
     port_type_fast: {{ interface['spanning_tree_portfast'] | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.spanning_tree_portfast) }}
@@ -45,6 +46,28 @@
     queuing_policy: {{ interface['queuing_policy'] | default("") }}
 {% endif %}
     enable_monitor: {{ interface['monitor'] | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.monitor) | lower }}
+    enable_storm_control: {{ interface['enable_storm_control'] | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.enable_storm_control) | lower }}
+{% if interface['enable_storm_control'] | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.enable_storm_control) %}
+    storm_control_action: {{ interface['storm_control_action'] | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.storm_control_action) | lower }}
+{% if interface['storm_control_broadcast_level_percent'] is defined %}
+    storm_control_broadcast_level_percent: "{{ '%0.2f' % interface['storm_control_broadcast_level_percent'] }}"
+{% endif %}
+{% if interface['storm_control_multicast_level_percent'] is defined %}
+    storm_control_multicast_level_percent: "{{ '%0.2f' % interface['storm_control_multicast_level_percent'] }}"
+{% endif %}
+{% if interface['storm_control_unicast_level_percent'] is defined %}
+    storm_control_unicast_level_percent: "{{ '%0.2f' % interface['storm_control_unicast_level_percent'] }}"
+{% endif %}
+{% if interface['storm_control_broadcast_level_pps'] is defined %}
+    storm_control_broadcast_level_pps: {{ interface['storm_control_broadcast_level_pps'] }}
+{% endif %}
+{% if interface['storm_control_multicast_level_pps'] is defined %}
+    storm_control_multicast_level_pps: {{ interface['storm_control_multicast_level_pps'] }}
+{% endif %}
+{% if interface['storm_control_unicast_level_pps'] is defined %}
+    storm_control_unicast_level_pps: {{ interface['storm_control_unicast_level_pps'] }}
+{% endif %}
+{% endif %}
     cmds: |2-
       {{ interface['freeform_config'] | default('') | indent(6, false) }}
 {% endif %}
--- a/roles/dtc/common/templates/ndfc_interfaces/ndfc_interface_trunk_po.j2
+++ b/roles/dtc/common/templates/ndfc_interfaces/ndfc_interface_trunk_po.j2
@@ -25,7 +25,7 @@
     enable_cdp: {{ interface['enable_cdp'] | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.enable_cdp) | lower }}
     mode: 'trunk'
     description: "{{ interface['description'] | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.description) }}"
-    copy_description: {{ interface['copy_description'] | default(defaults.vxlan.topology.switches.interfaces.topology_switch_access_po_interface.copy_description) | lower }}
+    copy_description: {{ interface['copy_description'] | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.copy_description) | lower }}
     mtu: {{ interface['mtu'] | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.mtu) }}
     speed: {{ interface['speed'] | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.speed) }}
     port_type_fast: {{ interface['spanning_tree_portfast'] | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.spanning_tree_portfast) }}
--- a/roles/dtc/common/templates/ndfc_interfaces/ndfc_interface_trunk_po.j2
+++ b/roles/dtc/common/templates/ndfc_interfaces/ndfc_interface_trunk_po.j2
@@ -47,6 +47,7 @@
     queuing_policy: {{ interface['queuing_policy'] | default("") }}
 {% endif %}
     enable_monitor: {{ interface['monitor'] | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.monitor) | lower }}
+{% if ndfc_version | cisco.nac_dc_vxlan.version_compare('12.4.1', '>=') %}
     enable_storm_control: {{ interface['enable_storm_control'] | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.enable_storm_control) | lower }}
 {% if interface['enable_storm_control'] | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.enable_storm_control) %}
     storm_control_action: {{ interface['storm_control_action'] | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.storm_control_action) | lower }}
@@ -67,6 +68,7 @@
 {% endif %}
 {% if interface['storm_control_unicast_level_pps'] is defined %}
     storm_control_unicast_level_pps: {{ interface['storm_control_unicast_level_pps'] }}
+{% endif %}
 {% endif %}
 {% endif %}
     cmds: |2-
--- a/roles/dtc/common/templates/ndfc_interfaces/ndfc_interface_vpc.j2
+++ b/roles/dtc/common/templates/ndfc_interfaces/ndfc_interface_vpc.j2
@@ -17,20 +17,21 @@
     - {{ switches[peer2]['mgmt_ip_address'] }}
   deploy: false
   profile:
-    admin_state: {{ switches[peer1].enabled | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.enabled) | lower }}
     mode: {{ switches[peer1].mode }}
     peer1_pcid: {{ switches[peer1].name | regex_replace('[^0-9]', '') }}
     peer2_pcid: {{ switches[peer2].name | regex_replace('[^0-9]', '') }}
-    port_type_fast: {{ switches[peer1].spanning_tree_portfast | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.spanning_tree_portfast) | lower }}
-    mtu: {{ switches[peer1].mtu | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.mtu) }}
-    speed: {{ switches[peer1].speed | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.speed) }}
-    pc_mode: "{{ switches[peer1].pc_mode | default(defaults.vxlan.topology.switches.interfaces.topology_switch_access_po_interface.pc_mode) }}"
     peer1_members: {{ switches[peer1].members | to_json }}
     peer2_members: {{ switches[peer2].members | to_json }}
     peer1_cmds: |2-
       {{ switches[peer1].freeform_config | default('') | indent(6, false) }}
     peer2_cmds: |2-
       {{ switches[peer2].freeform_config | default('') | indent(6, false)  }}
+{% if switches[peer1].mode == 'trunk' %}
+    admin_state: {{ switches[peer1].enabled | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.enabled) | lower }}
+    port_type_fast: {{ switches[peer1].spanning_tree_portfast | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.spanning_tree_portfast) | lower }}
+    mtu: {{ switches[peer1].mtu | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.mtu) }}
+    speed: {{ switches[peer1].speed | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.speed) }}
+    pc_mode: "{{ switches[peer1].pc_mode | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.pc_mode) }}"
     bpdu_guard: {{ switches[peer1].enable_bpdu_guard | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.enable_bpdu_guard) | lower }}
     disable_lacp_suspend_individual: {{ switches[peer1].disable_lacp_suspend_individual | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.disable_lacp_suspend_individual) | lower }}
     enable_lacp_vpc_convergence: {{ switches[peer1].enable_lacp_vpc_convergence | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.enable_lacp_vpc_convergence) | lower }}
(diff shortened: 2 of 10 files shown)


roles/dtc/common/templates/ndfc_interfaces/ndfc_interface_trunk_po.j2
```diff
@@ -25,6 +25,7 @@
     enable_cdp: {{ interface['enable_cdp'] | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.enable_cdp) | lower }}
     mode: 'trunk'
     description: "{{ interface['description'] | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.description) }}"
+    copy_description: {{ interface['copy_description'] | default(defaults.vxlan.topology.switches.interfaces.topology_switch_access_po_interface.copy_description) | lower }}
```
juburnet: In the trunk PO template, copy_description falls back to the wrong defaults key:

It should be:
copy_description: {{ interface['copy_description'] | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.copy_description) | lower }}

roles/validate/files/defaults.yml
```diff
@@ -235,6 +245,8 @@ factory_defaults:
             enable_bpdu_guard: true
             duplex: auto
             orphan_port: false
+            enable_storm_control: false
+            storm_control_action: default
```
juburnet, line 251: Look into defaults / schema naming issue 
defaults:
  vxlan:
    topology:
      switches:
        interfaces:
          topology_switch_dot1q_interface:   # <-- name in defaults
vs. in the schema diff:
topology_switch_dot1q_host_interface:       # <-- name in schema diff

juburnet (changes requested):

roles/validate/files/defaults.yml
juarocha: fixed on the schema

roles/dtc/common/templates/ndfc_interfaces/ndfc_interface_trunk_po.j2
juarocha: solved

roles/validate/files/rules/common/312_topology_interface_storm_control.py
```diff
@@ -0,0 +1,52 @@
...
+                        pps_ifaces.append(interface.get('name'))
+
+                if pct_ifaces and pps_ifaces:
+                    results.append(
+                        f"switch {switch.get('name')}: storm_control mode must be consistent "
+                        f"across all interfaces; percent-mode interfaces: {pct_ifaces}; "
+                        f"pps-mode interfaces: {pps_ifaces}"
+                    )
+
+        return results
```
juburnet, line 37: 🔴 Rule 312 — Wrong validation scope (per-switch instead of per-interface)
roles/validate/files/rules/common/312_topology_interface_storm_control.py
if pct_ifaces and pps_ifaces:
    results.append(...)
The rule fires when any interface on a switch uses percent-based levels AND any other interface on the same switch uses PPS-based levels. This is too broad. The actual NX-OS/NDFC constraint for the target hardware (Nexus 9000) is that within a single interface, you cannot mix percent and PPS for the same traffic type (e.g., storm_control_broadcast_level_percent and storm_control_broadcast_level_pps cannot both be set for broadcast on the same interface). But two different interfaces can independently use different modes.
A valid configuration that Rule 312 would incorrectly reject:
```
interfaces:
  - name: eth1/1
    enable_storm_control: true
    storm_control_broadcast_level_percent: 10  # percent mode
  - name: eth1/2
    enable_storm_control: true
    storm_control_broadcast_level_pps: 1000    # pps mode on a different interface
The correct logic should validate per-interface, per-traffic-type:
for interface in switch.get('interfaces'):
    for t in ['broadcast', 'multicast', 'unicast']:
        has_percent = interface.get(f'storm_control_{t}_level_percent') is not None
        has_pps = interface.get(f'storm_control_{t}_level_pps') is not None
        if has_percent and has_pps:
            results.append(
                f"switch {switch.get('name')}: interface {interface.get('name')}: "
                f"storm_control_{t} cannot specify both level_percent and level_pps"
            )
```
🟡 Rule 312 — Level fields not gated on enable_storm_control: true
The rule collects level keys from interfaces regardless of whether enable_storm_control is actually true for that interface. A user who defines level fields alongside enable_storm_control: false (misconfiguration or leftover data) won't get a validation error about the inconsistency, and those level values will silently be dropped from the rendered template (because the templates gate them on the enable flag). The rule should add:
```
if not interface.get('enable_storm_control', False):
    continue  # skip interfaces where storm control is disabled
```

roles/dtc/common/templates/ndfc_interfaces/ndfc_interface_vpc.j2
```diff
@@ -42,14 +43,75 @@
     qos_policy: {{ switches[peer1].qos_policy | default("") }}
     queuing_policy: {{ switches[peer1].queuing_policy | default("") }}
 {% endif %}
-{% if switches[peer1].mode == 'trunk' %}
+    copy_description: {{ switches[peer1].copy_description | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.copy_description) | lower }}
+    enable_storm_control: {{ switches[peer1].enable_storm_control | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.enable_storm_control) | lower }}
+{% if switches[peer1].enable_storm_control | default(defaults.vxlan.topology.switches.interfaces.topology_switch_trunk_po_interface.enable_storm_control) %}
```
juburnet, line 49: 🟡 ndfc_interface_vpc.j2 — copy_description and enable_storm_control placed outside the mode-conditional block for trunk VPCs
In the VPC template, copy_description (line 46) and enable_storm_control (line 47) appear after the QoS {% endif %} but they are still inside {% if switches[peer1].mode == 'trunk' %}. This is structurally correct — the qos {% endif %} on line 45 only closes the QoS sub-block. ✅ (No bug here, just confirming the indent intent is non-obvious.)
However: peer1_allowed_vlans and peer2_allowed_vlans (lines 69-70) are also inside the trunk block. If NDFC ever supports a vpc_peer_link mode, these fields won't be emitted. This is consistent with the original template behavior — noting for future extensibility awareness.

juburnet (changes requested):

roles/validate/files/rules/common/312_topology_interface_storm_control.py
juarocha: Resolved

juburnet (commented):
@juarocha  - Please check if this scenario would be an issue:

Rule 312 iterates switch.get('interfaces') but VPC peer configurations live under vxlan.topology.switches[].vpc_peers[].interfaces (or similar path). 

The VPC template now renders storm control for VPC interfaces. If a user sets both _percent and _pps on a VPC peer interface, Rule 312 would not catch it. 

Confirm whether VPC peer data is in a different path that needs a second traversal.

juarocha:
@juburnet just tested this, and it validates both sides of vPC. The rule is working for storm control is values but the we may be hitting #857

juburnet (approved):
Approved.
1. Consistent implementation across all 5+ interface templates — storm control blocks are identical in ndfc_interface_access.j2, ndfc_interface_access_po.j2, ndfc_interface_dot1q.j2, ndfc_interface_trunk.j2, ndfc_interface_trunk_po.j2, and ndfc_interface_vpc.j2.
2. Defaults properly seeded — defaults.yml updated for all affected interface types (dot1q, access, trunk, access_po, trunk_po). New copy_description: false default added wherever copy_description was introduced.
3. Validation rule 312 added — New 312_topology_interface_storm_control.py enforces the mutual exclusivity of percent vs. PPS. Good defensive validation.
4. vPC rule 305 bug fix is a real fix — The old logic (len(interfaces["interfaces"]) == 2) was incorrect in multi-fabric scenarios. The new per-pair iteration logic correctly scopes the consistency check to actual vPC peers.
5. data_model_key_check is self-contained — Rule 312 copies the utility method locally rather than inheriting from a base class inconsistently, matching the pattern of other rules in the codebase.
6. All CI passes — Clean across all Python/Ansible matrix combinations.

roles/validate/files/rules/common/312_topology_interface_storm_control.py
```diff
@@ -0,0 +1,51 @@
...
+            switches = data_model.get('vxlan').get('topology').get('switches')
+        else:
+            return results
+
+        for switch in switches:
+            check = cls.data_model_key_check(switch, ['interfaces'])
+            if 'interfaces' in check['keys_data']:
+                for interface in switch.get('interfaces'):
+
+                    if not interface.get('enable_storm_control', False):
```
mikewiebe: ### Medium: Disabled storm-control intent passes validation and is silently discarded

**Issue**

The source schema permits action and level fields alongside `enable_storm_control: false`. Rule 312 skips such an interface completely, and every changed Jinja template suppresses the action and levels when disabled. This reports validation success while discarding explicit user input; it also diverges from the new `ansible-dcnm` contract, which rejects dependent action/level values unless storm control is enabled.

**Evidence**

- Rule 312 returns early for disabled interfaces at [`if not interface.get('enable_storm_control', False): continue`](https://github.com/netascode/ansible-dc-vxlan/pull/855/files#diff-013c43f57a3c81890a0237a1ce8a51101143328e5af0be8d5558f3a0016b8216R24).
- The render path emits dependent fields only inside the enable gate, for example [the access port-channel block](https://github.com/netascode/ansible-dc-vxlan/pull/855/files#diff-eb0cea6817974c32d779e9e8e2760cedc42a0f337824c976308ec8e05df713c6R49).
- At source ref `7689d680aa5349b9a8175abe4e71cd4b9c883940`, the [schema makes the enable, action, and level keys independently optional](https://wwwin-github.cisco.com/netascode/nac-vxlan/blob/7689d680aa5349b9a8175abe4e71cd4b9c883940/schemas/schema.yaml#L466-L491).
- A practical example is a user disabling a previously enabled configuration but leaving its dependent values in the model:

  ```yaml
  vxlan:
    topology:
      switches:
        - name: leaf1
          interfaces:
            - name: Ethernet1/10
              mode: access
              enable_storm_control: false
              storm_control_action: shutdown
              storm_control_broadcast_level_percent: 20
  ```

  Rule 312 currently passes this model, and the template silently reduces the storm-control portion to:

  ```yaml
  enable_storm_control: false
  ```

  `storm_control_action` and `storm_control_broadcast_level_percent` disappear without an error. Passing the equivalent explicit profile to `ansible-dcnm` `develop` raises `storm-control action and levels require enable_storm_control: true`, matching its [validator contract](https://github.com/CiscoDevNet/ansible-dcnm/blob/b73d449a2a52e9ac9c166997ca8d20a508ff753f/plugins/modules/dcnm_interface.py#L2457-L2511).

**Existing PR overlap**

The silent-drop concern was explicitly raised in [discussion `r3677796609`](https://github.com/netascode/ansible-dc-vxlan/pull/855#discussion_r3677796609), then marked resolved after adding the skip. The current change fixes the thread's per-interface scope issue, but the skip preserves the distinct inconsistency that the same comment identified. Treat this as a previously raised concern whose resolution is incomplete.

**Impact**

Typos, stale intent, or an attempted disable-plus-reset can pass schema and semantic validation but never reach NDFC. The playbook result does not tell the operator that the action and levels were ignored.

**Suggested fix**

Implement the dependency check in Rule 312 before its current disabled-interface `continue`. Treat an absent enable flag as disabled because its modeled default is `false`; reject any configured percentage/PPS level and any non-default action, but continue allowing `storm_control_action: default`.

For example:

```python
level_keys = pct_keys | pps_keys

for switch in switches:
    for interface in switch.get("interfaces", []):
        enabled = interface.get("enable_storm_control", False)
        action = interface.get("storm_control_action")
        configured_levels = sorted(
            key for key in level_keys
            if interface.get(key) is not None
        )

        if not enabled:
            invalid_keys = list(configured_levels)
            if action not in (None, "default"):
                invalid_keys.append("storm_control_action")

            if invalid_keys:
                results.append(
                    f"switch {switch.get('name')} "
                    f"interface {interface.get('name')}: "
                    "storm-control action and level fields require "
                    "enable_storm_control: true; "
                    f"invalid keys: {sorted(invalid_keys)}"
                )
            continue

        pct_set = sorted(
            key for key in pct_keys
            if interface.get(key) is not None
        )
        pps_set = sorted(
            key for key in pps_keys
            if interface.get(key) is not None
        )

        if pct_set and pps_set:
            results.append(
                f"switch {switch.get('name')} "
                f"interface {interface.get('name')}: "
                "storm_control cannot mix percent and PPS "
                "on the same interface; "
                f"percent keys: {pct_set}; PPS keys: {pps_set}"
            )
```

Keep the Jinja enable gates as defense in depth. Add Rule 312 tests for disabled with no dependencies, disabled plus `default`, disabled plus `shutdown`/`trap`, disabled plus percent/PPS levels including zero, enabled percent-only, enabled PPS-only, and enabled mixed rate modes.

juarocha: Rule fixed

dacasti2mx (changes requested):
Rule 312 assumes storm control is disabled whenever ```enable_storm_control``` is not explicitly repeated on the interface, but, this setting can also be inherited from ```defaults.vxlan```

Since semantic validation runs before factory and custom defaults are merged, a valid interface that inherits ```enable_storm_control: true``` and only defines
its threshold is rejected.

Could Rule 312 validate the effective value using this precedence: 
 - explicit interface value, matching custom default, then factory default? 
 - It would also behelpful to add a test covering this inherited-default scenario.

For example, this should be valid because storm control is enabled through the access-interface defaults:
```yaml
defaults:
vxlan:
  topology:
    switches:
      interfaces:
        topology_switch_access_interface:
          enable_storm_control: true
```
```yaml
vxlan:
topology:
  switches:
    - name: leaf-1
      interfaces:
        - name: Ethernet1/1
          mode: access
          storm_control_broadcast_level_percent: 10
```
in this example, Rule 312 sees no ```enable_storm_control``` directly under Ethernet1/1, assumes false, and rejects the threshold even though the effective value inherited
from defaults is true.

juarocha:
Rule logic fixed. since defaults are merged after the validation we can't use them on the rule. The only validation is that percent-based and pps-based settings are not mixed on the same interface

juburnet:
@juarocha for Storm control, add ND 4.1 version requirement in templates as discussed.

juarocha:
> @juarocha for Storm control, add ND 4.1 version requirement in templates as discussed.

done

dacasti2mx (approved):
##  live validation
     
  **Env:** ND 4.1.1g / NDFC 12.4.1.321 · MCFG child fabric (FAB1) · targets NX-OSv (N9K-C9300v, 10.5(5))
  **Heads:** validated live at `318b9cd`; re-checked at `21d215e` (delta = templates-only version gate, 0 rule changes)
  **Result:** **LGTM** — features validated end-to-end; the "no config when not applicable" behaviors are by-design (see 
  Notes).

  ### Negative / validation
  | # | Model | Expected | Result |
  |---|---|---|---|
  | N1 | percent + pps on same intf | Rule 312 FAIL | ✅ 312 FAIL |
  | N2 | `enable:false` + threshold | accepted; threshold not applied (storm-control off) | ✅ nac-validate passes; render
  omits it (by design) |
  | N3 | vPC peers differ | Rule 305 FAIL | ✅ 305 FAIL |
  | N4 | same `vpc_id` on 2 pairs | PASS | ✅ PASS (pairing fix) |

  ### Physical interface (subject `Eth1/7`, control `Eth1/8` untouched)
  | Stage | Change | Result (NX-OS) |
  |---|---|---|
  | L0 | access, no storm-control | ✅ access, no storm-control |
  | L1 | enable + shutdown + bcast 5.00% | ✅ `storm-control broadcast level 5.00` + `action shutdown` |
  | L2 | rerun | ✅ idempotent |
  | L3 | 20.00% + trap + mcast 10.00% | ✅ updated cleanly (no stale values) |
  | L4 | percent → pps 1000 | ✅ `storm-control broadcast level pps 1000` |
  | L5 | disable | ✅ storm-control fully removed |
  | L6 | trunk + 10.00% | ✅ trunk + storm-control |
  | L7 | dot1q + 10.00% | ✅ `mode dot1q-tunnel` + storm-control |
  | L8 | restore | ✅ clean |
  | inherited | default enable=true, no explicit enable | ✅ threshold renders (inheritance works) |

  ### Port-channels + copy_description
  | Stage | Change | Result |
  |---|---|---|
  | P1 | access PO + 5% + `copy_description:true` | ✅ PO storm-control; description copied to members |
  | P2 | verify | ✅ member desc == PO desc |
  | P3 | update description | ✅ propagated to members |
  | P4 | percent → pps | ✅ clean transition |
  | P5 | `copy_description:false` | ✅ stops copying; previously-copied member desc retained (by design) |
  | P6 | trunk PO + storm-control | ✅ (see Notes: NX-OS box-wide %/pps constraint) |
  | P7 | routed PO + `copy_description`, no storm-control | ✅ copy_description works; storm-control correctly absent |
  | P8 | removal | ✅ requires `role_remove` (merged create is additive) |
  | P9 | rerun | ✅ idempotent |

  ### vPC / ToR pair
  | # | Case | Result |
  |---|---|---|
  | D1–D3/D5 | storm-control + copy_description on vPC PO, both peers | ✅ propagated to both peers + members; Rule 305 PASS |
  | D4 | pps on vPC | ⏭️  by equivalence (pps validated in L4/P4) |
  | D6 | update | ✅ propagated to both |
  | D7 | removal | ✅ reverted cleanly |
  | D8 | rerun | ✅ idempotent |
  | D9 | peers differ | ✅ Rule 305 FAIL before mutation |
  | D10 | percent + pps | ✅ Rule 312 FAIL |
  | D11 | same `vpc_id` on 2 pairs | ✅ Rule 305 PASS (pairing fix) |

PR #860, opened 2026-07-31
fix ipv4 subnet mask with manual allocation ebgp

### LINKED ISSUE #861: SUBNET_TARGET_MASK missing from eBGP VXLAN fabric payload when manual_underlay_allocation is enabled
## Problem

When creating/updating an eBGP VXLAN fabric with `manual_underlay_allocation: true`, the `SUBNET_TARGET_MASK` parameter is not included in the NDFC fabric payload. NDFC then uses its default value instead of the user-configured `vxlan.underlay.ipv4.subnet_mask`.

## Root Cause

In `roles/dtc/common/templates/ndfc_fabric/ebgp_vxlan_fabric/general/ebgp_vxlan_fabric_general.j2`, `SUBNET_TARGET_MASK` was nested inside the `manual_underlay_allocation == false` guard:

```jinja2
{% if not manual_underlay_allocation %}
{% if not enable_ipv6_underlay %}
  SUBNET_TARGET_MASK: "{{ subnet_mask }}"
  SUBNET_RANGE: {{ subnet_range }}
{% endif %}
{% endif %}
```

This incorrectly omits `SUBNET_TARGET_MASK` when manual allocation is enabled. However, NDFC requires this parameter regardless of allocation mode — it defines the point-to-point link prefix length used for all underlay links.

The only valid reason to omit `SUBNET_TARGET_MASK` is when `enable_ipv6_underlay` is true (IPv6 uses different parameters).

## Impact

- eBGP VXLAN fabrics with `manual_underlay_allocation: true` get NDFC's default subnet mask instead of the configured value
- May cause underlay link IP addressing mismatches between data model and NDFC intent

## Fix

Move `SUBNET_TARGET_MASK` outside the `manual_underlay_allocation` guard, keeping only the `enable_ipv6_underlay` condition. `SUBNET_RANGE` remains correctly gated by manual allocation (since ranges are irrelevant when IPs are manually assigned).

## Affected Versions

Any ND version with eBGP VXLAN fabric using `manual_underlay_allocation: true`.

ccoueffe:
## Related Issue(s)

Fixes missing `SUBNET_TARGET_MASK` in eBGP VXLAN fabric payload when `manual_underlay_allocation` is enabled.

## Related Collection Role

* [x] other — `roles/dtc/common/templates/ndfc_fabric/ebgp_vxlan_fabric/general/`

## Related Data Model Element

* [x] vxlan.underlay

## Proposed Changes

The `SUBNET_TARGET_MASK` parameter was nested inside the `manual_underlay_allocation == false` guard in the eBGP fabric template. This meant that when using manual underlay allocation, the subnet mask was not sent to NDFC — causing fabric creation/update to use NDFC defaults instead of the user-defined value.

`SUBNET_TARGET_MASK` is required by NDFC regardless of whether IP allocation is manual or automatic — it defines the point-to-point link prefix length for the underlay. The only condition where it should be omitted is when IPv6 underlay is enabled (since IPv6 uses a different parameter).

**Fix:** Moved `SUBNET_TARGET_MASK` outside the `manual_underlay_allocation` guard, keeping only the `enable_ipv6_underlay` condition:

| Condition | Before (broken) | After (fixed) |
|---|---|---|
| `manual=false`, `ipv6=false` | ✅ Rendered | ✅ Rendered |
| `manual=true`, `ipv6=false` | ❌ Missing | ✅ Rendered |
| `manual=false`, `ipv6=true` | ❌ Omitted | ❌ Omitted |
| `manual=true`, `ipv6=true` | ❌ Omitted | ❌ Omitted |

## Test Notes

Tested with eBGP VXLAN fabric using `manual_underlay_allocation: true` and `subnet_mask: 30`. Verified:
- `SUBNET_TARGET_MASK: "30"` now appears in the fabric payload ✅
- `SUBNET_RANGE` remains correctly omitted for manual allocation ✅
- IPv6 underlay correctly omits `SUBNET_TARGET_MASK` ✅

ND 4.1+

## Checklist

### DIFF
--- a/roles/dtc/common/templates/ndfc_fabric/ebgp_vxlan_fabric/general/ebgp_vxlan_fabric_general.j2
+++ b/roles/dtc/common/templates/ndfc_fabric/ebgp_vxlan_fabric/general/ebgp_vxlan_fabric_general.j2
@@ -14,9 +14,11 @@
 {% endif %}
   UNDERLAY_IS_V6: {{ vxlan.underlay.general.enable_ipv6_underlay | default(defaults.vxlan.underlay.general.enable_ipv6_underlay) | title }}
   STATIC_UNDERLAY_IP_ALLOC: {{ vxlan.underlay.general.manual_underlay_allocation | default(defaults.vxlan.underlay.general.manual_underlay_allocation) | title }}
-{% if not (vxlan.underlay.general.manual_underlay_allocation | default(defaults.vxlan.underlay.general.manual_underlay_allocation) | ansible.builtin.bool) %}
 {% if not (vxlan.underlay.general.enable_ipv6_underlay | default(defaults.vxlan.underlay.general.enable_ipv6_underlay) | ansible.builtin.bool) %}
   SUBNET_TARGET_MASK: "{{ vxlan.underlay.ipv4.subnet_mask | default(defaults.vxlan.underlay.ipv4.subnet_mask) }}"
+{% endif %}
+{% if not (vxlan.underlay.general.manual_underlay_allocation | default(defaults.vxlan.underlay.general.manual_underlay_allocation) | ansible.builtin.bool) %}
+{% if not (vxlan.underlay.general.enable_ipv6_underlay | default(defaults.vxlan.underlay.general.enable_ipv6_underlay) | ansible.builtin.bool) %}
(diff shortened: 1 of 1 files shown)


juburnet (approved):
Approved.

marehler (approved):
Looks good to me.

mikewiebe (approved):

PR #874, opened 2026-08-18
Adding qos parameters to validate rule

### LINKED ISSUE #873: Validation  for QoS parameters on vpc is not done
### Ansible Version

```shell
ansible [core 2.18.18]
```

```shell
Collection        Version   
----------------- ----------
ansible.netcommon 8.6.1     
ansible.utils     6.0.3     
cisco.dcnm        3.12.0-dev
```

```shell
4.1
```

### Cisco NX-OS Version

```shell
10.6(2)
```

cisco.nac_dc_vxlan.validate

vxlan.topology

### Expected Behavior

```shell
role validate should send an error that qos parameters are not the same on vPC interfaces
```

### Actual Behavior

```shell
validate doesn't check this parameters
```

### Ansible Playbook

```yaml
ansible-playbook -i inventory.yaml vxlan.yaml -l fabric1 --tags=role_validate
```

### Data Model

```yaml
---
vxlan:
  topology:
    switches:
      - name: fabric1-l1
        interfaces:
          - name: port-channel11
            mode: trunk
            pc_mode: active
            vpc_id: 11  # VPC Port-channel
            # enable_qos: true                        <<<<<<<< No QoS defined
            qos_policy: port_qos_policy
            queuing_policy: port_queuing_policy
      - name: fabric1-l2
        interfaces:
          - name: port-channel11
            mode: trunk
            pc_mode: active
            vpc_id: 11  # VPC Port-channel (matches fabric1-l1)
            members:
              - eth1/18
            enable_qos: true
            qos_policy: port_qos_policy
            queuing_policy: port_queuing_policy
```

### Steps to Reproduce

```shell
run the playbook
```

### Relevant Debug Output

```shell

```

juarocha:
## Related Issue(s)

Fixes #873 

## Related Collection Role

* [X] cisco.nac_dc_vxlan.validate
* [ ] other

## Related Data Model Element

* [ ] vxlan.fabric
* [ ] vxlan.global
* [X] vxlan.topology
* [ ] vxlan.underlay
* [ ] vxlan.overlay
* [ ] vxlan.policy
* [ ] vxlan.multisite
* [ ] defaults.vxlan
* [ ] other

## Proposed Changes
Adding **enable_qos**, **qos_policy** and **queuing_policy** to validate rules

4.1

## Checklist

* [X] Latest commit is rebased from develop with merge conflicts resolved
* [X] New or updates to documentation has been made accordingly
* [X] Assigned the proper reviewers

### DIFF
--- a/roles/validate/files/rules/common/305_topology_switch_interfaces_vpc.py
+++ b/roles/validate/files/rules/common/305_topology_switch_interfaces_vpc.py
@@ -28,6 +28,9 @@ class Rule:
             "storm_control_multicast_level_pps",
             "storm_control_unicast_level_percent",
             "storm_control_unicast_level_pps",
+            "enable_qos",
+            "qos_policy",
+            "queuing_policy",
         ]
         vpc_peers_list = cls.get_vpc_peers(data_model)
         # Check if fabric topology switches are defined


dacasti2mx (changes requested):
I found one valid scenario that is rejected because the comparison does not consider the effective default value.

The problem is the comparison between an omitted `enable_qos` value and an explicit `false`.

For example:

- Peer 1: `enable_qos` is omitted.
- Peer 2: `enable_qos: false`.

Both configurations have the same effective result because the default value of `enable_qos` is `false`.

However, Rule 305 compares the raw values:

- An omitted value becomes `None`.
- An explicit value remains `False`.

Because `None != False`, the validation rejects a model where QoS is effectively disabled on both peers.

The error returned was:

```text
Semantic error, rule 305: ... have different enable_qos values
```
Before comparing the vPC peers, `enable_qos` should be normalized using its effective default value.

I recommend addressing the default-value comparison before merging.

juarocha:
Defaults are merged after the validation rules. this cannot be checked with the current collection workflow

dacasti2mx:
> Defaults are merged after the validation rules. this cannot be checked with the current collection workflow

Problem found will be solved in another PR. for #874 code and tests LGMT.

dacasti2mx (approved):
LGTM, reported problem will be addressed in another PR.
#874 code and testing works ok.

juburnet (approved):
The fix is correct and minimal. Adding enable_qos, qos_policy, and queuing_policy to vpc_params_to_match in rule 305 is precisely the right place — that list is the single source of truth for which fields must be symmetric across vPC peer interfaces. The existing comparison logic (a[param] != b[param]) handles these new fields correctly without any modification.

Correct fix location	- vpc_params_to_match is the exact right list — adding fields here extends enforcement with zero code duplication
Issue example is caught	- The case in #873 (enable_qos absent on one peer, true on the other) produces None != True → validation error fires correctly
Template alignment	- The VPC interface template (ndfc_interface_vpc.j2) already supports all three fields; the validate rule was simply lagging behind
No false negatives	- interface.get(param) returns None for absent keys, and None != True and None != False both correctly trigger the mismatch error

mikewiebe (approved):
LGTM

PR #870, opened 2026-08-12
fix: send default OSPF auth key ID (127) when authentication is disabled

ccoueffe:
Fixes #869

## Related Collection Role

- [ ] cisco.nac_dc_vxlan.validate
- [x] cisco.nac_dc_vxlan.dtc.create
- [ ] cisco.nac_dc_vxlan.dtc.deploy
- [ ] cisco.nac_dc_vxlan.dtc.remove
- [ ] other

## Related Data Model Element

- [ ] vxlan.fabric
- [ ] vxlan.global
- [ ] vxlan.topology
- [x] vxlan.underlay
- [ ] vxlan.overlay
- [ ] vxlan.overlay_extensions
- [ ] vxlan.policy
- [ ] vxlan.multisite
- [ ] defaults.vxlan
- [ ] other

## Proposed Changes

When `ospf.authentication_enable: false`, NDFC rejects any `OSPF_AUTH_KEY_ID` value other than empty or `127`. Previously the template always sent the user-provided key ID, causing fabric update failures when auth was disabled but a key ID was defined in the data model.

### Template fix (`dc_vxlan_fabric_protocols.j2`)

| Condition | Before | After |
|-----------|--------|-------|
| `authentication_enable: true` | Sends user key ID + key | Sends user key ID + key (unchanged) |
| `authentication_enable: false` | Sends user key ID (fails if != 127) | Sends default key ID (`127` from defaults) |

The `else` branch now explicitly sends the default `OSPF_AUTH_KEY_ID` (127) when authentication is disabled. This is safe because:
- `127` is the NDFC default value for `OSPF_AUTH_KEY_ID` in the `dcnm_fabric` module
- `127` is accepted by NDFC in all scenarios (fresh fabric, previously enabled then disabled)
- `OSPF_AUTH_KEY` is not sent when auth is disabled (omitted from payload)

### Data Model Example

```yaml
vxlan:
  underlay:
    ospf:
      area_id: 0.0.0.0
      authentication_enable: false    # auth disabled
      authentication_key_id: 10       # ignored — template sends 127 instead
      authentication_key: my_key      # ignored — not sent to NDFC
```

## Test Notes

Tested with:
- Fresh fabric with auth disabled → ✅ deploys cleanly
- Fabric with auth previously enabled then disabled → ✅ `OSPF_AUTH_KEY_ID=127` accepted
- Fabric with auth enabled → ✅ user key ID sent as before

## Cisco Nexus Dashboard Version

12.4.1

## Checklist

[...]

PR #880, opened 2026-08-27
fix: handle missing response key in ndfc_get_switch_policy_using_desc

juburnet:
## Summary

Defensive robustness improvement for `ndfc_get_switch_policy_using_desc`.

## Problem

On `sac-nd-41-ca/273`, the remove stage failed with:

```
REMOVE [nac-fabric2-ca] Discovering and removing unmanaged policies across 4 switch(es)
fatal: [nac-fabric2-ca]: FAILED! => {"changed": false, "msg": "Remove resources failed: 'response'"}
```

`ndfc_get_switch_policy_using_desc` (line 233) does a bare key access `policy_data["response"]["DATA"]`. When `dcnm_rest` returns without a `"response"` key (e.g. a transient NDFC API failure or connection error), this raises `KeyError: 'response'`, which propagates as the cryptic `Remove resources failed: 'response'` message.

`ndfc_get_switch_policy_using_template` already had this fixed (line 184) with the defensive `.get()` pattern. `ndfc_get_switch_policy_using_desc` was missed.

## Root Cause

Build 273 failed due to a **transient NDFC API failure** — `dcnm_rest` returned without a `"response"` key. Builds 270–271 on the same code path passed without issue. `dcnm_rest.py` in `cisco.dcnm` has not changed since March 2026. This is a pre-existing defensive gap, not a regression introduced by today's merges (#870, #874, #775).

**Recommended action for `sac-nd-41-ca`**: rerun the pipeline. This PR does not prevent failure if NDFC is unresponsive — it ensures the failure surfaces with a clear, actionable error message instead of a cryptic `KeyError`.

## Fix

Apply the same pattern already used in `ndfc_get_switch_policy_using_template`:

```python
# Before
policy_match = [
    item for item in policy_data["response"]["DATA"]
    ...
]

# After
policy_response = policy_data.get("response") or policy_data.get("msg") or {}
return_code = policy_response.get("RETURN_CODE")
if return_code != 200:
    raise Exception(f"Policy lookup failed for switch {switch_serial_number}: {policy_response}")

policy_match = [
    item for item in policy_response.get("DATA", [])
    ...
]
```

## Related

- Observed on `sac-nd-41-ca` Run 273
- Pre-existing defensive gap — not caused by today's merges

## Checklist

- [x] Single-file fix — no schema, template, or default changes
- [x] Follows existing defensive `.get()` pattern from `ndfc_get_switch_policy_using_template`
- [x] Replaces cryptic `KeyError` with a descriptive exception message for easier future debugging

### DIFF
--- a/plugins/plugin_utils/helper_functions.py
+++ b/plugins/plugin_utils/helper_functions.py
@@ -229,8 +229,13 @@ def ndfc_get_switch_policy_using_desc(self, task_vars, tmp, switch_serial_number
     """
     policy_data = ndfc_get_switch_policy(self, task_vars, tmp, switch_serial_number)
 
+    policy_response = policy_data.get("response") or policy_data.get("msg") or {}
+    return_code = policy_response.get("RETURN_CODE")
+    if return_code != 200:
+        raise Exception(f"Policy lookup failed for switch {switch_serial_number}: {policy_response}")
+
     policy_match = [
-        item for item in policy_data["response"]["DATA"]
+        item for item in policy_response.get("DATA", [])
         if item.get("description", None) and item.get("description", None).startswith(prefix) and item["source"] == ""
     ]
 
--- a/plugins/plugin_utils/helper_functions.py
+++ b/plugins/plugin_utils/helper_functions.py
@@ -229,7 +229,11 @@ def ndfc_get_switch_policy_using_desc(self, task_vars, tmp, switch_serial_number
     """
     policy_data = ndfc_get_switch_policy(self, task_vars, tmp, switch_serial_number)
 
-    policy_response = policy_data.get("response") or policy_data.get("msg") or {}
+    # Fallback to an empty dict if the result is not a dictionary
+    policy_response = policy_data.get("response") or policy_data.get("msg")
+    if not isinstance(policy_response, dict):
+        policy_response = {}
+
     return_code = policy_response.get("RETURN_CODE")
     if return_code != 200:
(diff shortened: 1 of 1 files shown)


marehler (approved):
Looks good to me. Correct root-cause analysis, minimal single-file change, mirrors an established pattern. However I don't have a chance to reproduce the issue and test the fix.

plugins/plugin_utils/helper_functions.py
```diff
@@ -229,8 +229,13 @@ def ndfc_get_switch_policy_using_desc(self, task_vars, tmp, switch_serial_number
     """
     policy_data = ndfc_get_switch_policy(self, task_vars, tmp, switch_serial_number)
 
+    policy_response = policy_data.get("response") or policy_data.get("msg") or {}
```
mikewiebe: There is still a corner case that could cause a crash here.  

If `policy_data.get("response")` returns a string or a boolean instead of a dictionary (for example, if the API returns a text error message like `"Internal Server Error"`), the code will crash.

This version is safer:

```python
# Fallback to an empty dict if the result is not a dictionary
policy_response = policy_data.get("response") or policy_data.get("msg")
if not isinstance(policy_response, dict):
    policy_response = {}

return_code = policy_response.get("RETURN_CODE")
if return_code != 200:
    raise Exception(f"Policy lookup failed for switch {switch_serial_number}: {policy_response}")
```

juburnet: updated

mikewiebe (approved):

